PR #3, opened 2013-10-31
PKCS11 token support

erwint:
this patch enables using a custom provider, most notably the PKCS11 provider which is used for signing using a key stored on a Crypto Token.

Example:

java net.jsign.PESignerCLI -providerClass sun.security.pkcs11.SunPKCS11 -providerArg my.conf --storetype PKCS11 --storepass 1234 --t http://timestamp.comodoca.com/authenticode --alias XYZ file

Unfortuantely it's a bit complicated to provide an automated test case, as you either need an actual token when running the test or implement an own dummy-Provider.

### DIFF
--- a/jsign-cli/src/test/java/net/jsign/PESignerCLITest.java
+++ b/jsign-cli/src/test/java/net/jsign/PESignerCLITest.java
@@ -18,13 +18,12 @@ package net.jsign;
 
 import java.io.File;
 import java.lang.reflect.InvocationTargetException;
+import java.security.InvalidParameterException;
 import java.security.Permission;
 import java.security.ProviderException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import io.netty.handler.codec.http.HttpRequest;
-import junit.framework.TestCase;
 import org.apache.commons.cli.ParseException;
 import org.apache.commons.io.FileUtils;
 import org.bouncycastle.cms.CMSSignedData;
@@ -34,6 +33,8 @@ import org.littleshoot.proxy.HttpProxyServer;
 import org.littleshoot.proxy.ProxyAuthenticator;
 import org.littleshoot.proxy.impl.DefaultHttpProxyServer;
 
+import io.netty.handler.codec.http.HttpRequest;
+import junit.framework.TestCase;
 import net.jsign.pe.PEFile;
 
 public class PESignerCLITest extends TestCase {
@@ -515,7 +516,8 @@ public class PESignerCLITest extends TestCase {
             fail("No exception thrown");
         } catch (SignerException e) {
             // expected
-            assertTrue(e.getCause().getCause() instanceof ProviderException);
+            assertTrue(e.getCause().getCause() instanceof ProviderException // JDK < 9
+            		|| e.getCause().getCause() instanceof InvalidParameterException); // JDK 9+
         }
     }
 
--- a/jsign-cli/src/test/java/net/jsign/PESignerCLITest.java
+++ b/jsign-cli/src/test/java/net/jsign/PESignerCLITest.java
@@ -17,6 +17,7 @@
 package net.jsign;
 
 import java.io.File;
+import java.lang.reflect.InvocationTargetException;
 import java.security.Permission;
 import java.security.ProviderException;
 import java.util.List;
@@ -40,7 +41,7 @@ public class PESignerCLITest extends TestCase {
     private PESignerCLI cli;
     private File sourceFile = new File("target/test-classes/wineyes.exe");
     private File targetFile = new File("target/test-classes/wineyes-signed-with-cli.exe");
-    
+
     private String keystore = "keystore.jks";
     private String alias    = "test";
     private String keypass  = "password";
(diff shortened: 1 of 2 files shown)


ebourg:
@erwint Erwin if you are still around, out of curiosity what PKCS11 token did you use with this modification?

I'm looking into this token stuff, I think I'll reuse the `keystore` parameter to specify either the SunPKCS11 configuration file, or the name of the provider. I assume most PKCS11 tokens are usable with the SunPKCS11 provider, so adding a `providerClass` isn't essential. The syntax would look like this:

    pesign --storetype PKCS11 --keystore token.cfg --storepass secret --alias xyz file

and when the provider is configured in `java.security`:

    pesign --storetype PKCS11 --keystore SunPKCS11-foo --storepass secret --alias xyz file

erwint:
We were using it with different tokens, e.g. first an ASEKey and then a Nitrokey Pro (which handles 4096 Bits keys)

Depending on a configured provider in ```java.security``` is not a good idea, as you want a setup that does not require to modify the system's java installation (which you would need to repeat after every update). But as long it's just a default you can override it I don't see a problem.

Also, you could have a look at our current branch which is in production for some time now and contains some more changes -> https://github.com/GEBIT/jsign/commits/master

ebourg:
Thank you for the information. I've merged the TSA retry feature and I'm now working on PKCS11 support.

ebourg:
PKCS#11 support is now implemented, this is rather experimental though since I don't have access to a crypto token to test the implementation.

@erwint Could you give it a try and tell me if it works please?

erwint:
There are some issues left
* You need to be able to provide a certfile, as the Certificate (chain) is not always available from the Token (some hold only the pubkey)
* The providerArg need to be passed with the full path, not just the name
* The shaded jar did not contain the bouncycastle implementation, I disabled minimizing to try it out. The Jar is not much too big, though. Probably just some missing setting

With these changes I can sign using a Nitrokey Pro token

```
diff --git a/jsign-core/src/main/java/net/jsign/PESignerHelper.java b/jsign-core/src/main/java/net/jsign/PESignerHelper.java
index 0a7d097..3e37205 100644
--- a/jsign-core/src/main/java/net/jsign/PESignerHelper.java
+++ b/jsign-core/src/main/java/net/jsign/PESignerHelper.java
@@ -247,7 +247,7 @@ class PESignerHelper {
         if (keystore == null && keyfile == null && certfile == null) {
             throw new SignerException("keystore " + parameterName + ", or keyfile and certfile " + parameterName + "s must be set");
         }
-        if (keystore != null && (keyfile != null || certfile != null)) {
+        if (keystore != null && !"PKCS11".equals(storetype) && (keyfile != null || certfile != null)) {
             throw new SignerException("keystore " + parameterName + " can't be mixed with keyfile or certfile");
         }

@@ -255,7 +255,7 @@ class PESignerHelper {
         if ("PKCS11".equals(storetype)) {
             // the keystore parameter is either the provider name or the SunPKCS11 configuration file
             if (keystore != null && keystore.exists()) {
-                provider = new SunPKCS11(keystore.getName());
+                provider = new SunPKCS11(keystore.getPath());
             } else if (keystore != null && keystore.getName().startsWith("SunPKCS11-")) {
                 provider = Security.getProvider(keystore.getName());
                 if (provider == null) {
diff --git a/jsign/pom.xml b/jsign/pom.xml
index 8a99e83..4cf4a62 100644
--- a/jsign/pom.xml
+++ b/jsign/pom.xml
@@ -60,7 +60,7 @@
               <goal>shade</goal>
             </goals>
             <configuration>
-              <minimizeJar>true</minimizeJar>
+              <minimizeJar>false</minimizeJar>
               <filters>
                 <filter>
                   <artifact>*:*</artifact>
```

ebourg:
I fixed the all-in-one jar and the path to the SunPKC11 configuration. Regarding the certificate, are you sure it isn't stored with the private key on the token? Otherwise an exception would be thrown right after loading the keystore:

https://github.com/ebourg/jsign/blob/89a40551/jsign-core/src/main/java/net/jsign/PESignerHelper.java#L282

erwint:
Yes, I'm absoultely sure. Not all tokens support it, also see documentation of jarsigne:

``` 
-certchain file
Specifies the certificate chain to be used when the certificate chain associated with the private key of the keystore entry that is addressed by the alias specified on the command line is not complete. This can happen when the keystore is located on a hardware token where there is not enough capacity to hold a complete certificate chain. The file can be a sequence of concatenated X.509 certificates, or a single PKCS#7 formatted data block, either in binary encoding format or in printable encoding format (also known as Base64 encoding) as defined by the Internet RFC 1421 standard. See Internet RFC 1421 Certificate Encoding Standard and http://tools.ietf.org/html/rfc1421.```

ebourg:
Ok, but since no exception was thrown in your case I suspect there was at least one certificate on the token. That would mean that the token can hold the private key and the final certificate with the public key, but not the certificate chain of the issuer. So jsign would have to combine the certificate on the token with the chain from the file to get the full chain.

erwint:
yes, in that case you have the private key and public key with certificate, but not the chain on the token

ebourg:
@erwint I have eventually merged your latest changes, thanks again. This will be part of Jsign 3.1

PR #4, opened 2013-11-29
Improve EXE signing to be compatible with Launch4J on JDK 6

solerman:
While trying to use JSign with current Launch4j, java 6 doesn't accepts the generated package anymore. Another thread for one native signer suggested the fix which I implemented: https://sourceforge.net/p/launch4j/discussion/332683/thread/4e9e2558/


ebourg:
I'm tempted to close this issue as wontfix for two reasons:
- This looks like a flaw in the way launch4j loads the jar. launch4j should write the exact size of the jar somewhere in the executable and load only this amount of data as the jar instead of assuming that the end of the file is always a jar.
- This only happens with Java 6 which has been EOLed 3 years ago. Desktop users are likely to use more recent JREs now (my own stats for the desktop application I distribute show Java 6 under 15% and that's mostly OS X users).

solerman:
> This looks like a flaw in the way launch4j loads the jar. launch4j should write the exact size of the jar somewhere in the executable and load only this amount of data as the jar instead of assuming that the end of the file is always a jar.

I don't think jsign can propertly solve this in the current launch4j design, closing

PR #9, opened 2015-02-04
Simple support for HTTP proxies

szediwy:
This adds simple support for HTTP proxies. The proxy can be configured in ANT task properties or over command line arguments.


ebourg:
Thank you Michael. What about using a `ProxySelector` and setting a default `Authenticator` instead? This would apply to all http requests and avoid passing the proxy settings to the methods. I think we should also use the system proxy by default (by setting the system property `java.net.useSystemProxies` to `true`). If you could rework the patch that would great, otherwise I'll look at this when I have some time. Bonus point for tab-less indentation :)

szediwy:
I did not automatically set java.net.useSystemProxies to true in this version.

ebourg:
I'm going to merge only the CLI part, Ant already has a mechanism to specify the proxy configuration:

https://ant.apache.org/manual/proxy.html

PR #10, opened 2015-02-05
JSign Maven Plug-In

szediwy:
Instead of calling the ANT task in maven build it would be nice to have a maven plug-in to call.


ebourg:
Very nice, thank you! I can't merge it now because I have other changes pending, but I'll definitely add that. The only downside is the duplication of the parameter validation logic, I'll have to think how to merge it with the validation of the Ant task.

ebourg:
Another implementation: https://github.com/arxes-tolina/jsign-maven-plugin

PR #12, opened 2015-03-10
Setting correct snapshot version.

netmackan:
I noticed that the current version number used in pom.xml in the master branch is still 1.2. With Maven conventions, this should probably be changed to 1.3-SNAPSHOT and then changed to 1.3 when the 1.3 release is tagged.

Cheers,
Markus


ebourg:
You are right, thank you.

PR #13, opened 2015-03-10
Upgrading BouncyCastle version to 1.49.

netmackan:
In my project (signserver.org) we use BouncyCastle version 1.49 and Jsign seems to work fine with that version too as long as it was compiled against it. Feel free to ignore/close this PR in case you don't want to upgrade BC at this point.

Cheers,
Markus

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,13 +45,13 @@
     <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcprov-jdk15on</artifactId>
-      <version>1.48</version>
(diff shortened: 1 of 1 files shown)


ebourg:
The upgrade to BC 1.51 is likely to be more tricky though...

netmackan:
True.
I expect us to upgrade SignServer to BC 1.51 (or maybe 1.52) within a few months and then we will also have to upgrade Jsign so unless you are in hurry you can wait for my pull request after that.

Cheers,
Markus

PR #14, opened 2015-03-10
Removing shade plugin.

netmackan:
I have had issues with the maven-shade-plugin as it creates a jar with the dependencies included. My project already bundles the dependencies with the application so it does not make sense for us to also have them duplicated in the jsign jar.
In my fork I simply removed the plugin from pom.xml, however maybe there is some other option to have it conditionally included (?).

Not that important but NetBeans IDE also reports this as an issue:
"When the final artifact jar contains classes not originating in current project, NetBeans internal compiler cannot use the sources of the project for compilation. Then changes done in project's source code only appears in depending projects when project is recompiled. Also applies to features like Refactoring which will not be able to find usages in depending projects."

Cheers,
Markus

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,36 +86,6 @@
         </configuration>
       </plugin>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-shade-plugin</artifactId>
-        <version>1.6</version>
-        <executions>
-          <execution>
-            <phase>package</phase>
-            <goals>
-              <goal>shade</goal>
-            </goals>
-            <configuration>
-              <minimizeJar>true</minimizeJar>
-              <filters>
-                <filter>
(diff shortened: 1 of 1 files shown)


ebourg:
I'd like to keep a standalone jar to be used with Ant, it's more convenient to drop a single jar in a directory than a set of dependencies. That said we could also tweak the plugin configuration to relocate the classes and avoid conflicts.

ebourg:
I have relocated the dependencies included in the jar in 51e2ab3, so you shouldn't have conflicts with your project dependencies now. I think I'll later split jsign into modules (jsign-core, jsign-cli, jsign-ant, jsign-maven-plugin) and only the cli and ant jars will include the dependencies. jsign-core will be the right dependency for reusing jsign into other projects.

netmackan:
Created #25 so this can closed.

PR #16, opened 2015-03-10
Adding support for providing a custom Timestamper implementation.

netmackan:
The following patch adds support for providing the PESigner your own implementation of the timestamp client (Timestamper). 

Cheers,
Markus

### DIFF
--- a/src/main/java/net/jsign/PESigner.java
+++ b/src/main/java/net/jsign/PESigner.java
@@ -82,6 +82,7 @@ public class PESigner {
     private boolean timestamping = true;
     private TimestampingMode tsmode = TimestampingMode.AUTHENTICODE;
     private String tsaurlOverride;
+    private Timestamper timestamper;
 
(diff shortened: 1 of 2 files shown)


ebourg:
For my understanding, could you explain a bit why you have to set a custom timestamper?

netmackan:
The main reason is that we give the user the option to use our internal timestamp-signer. In that case we do not use HTTP but instead our internal APIs (direct EJB calls within the application server).

Other reasons could be if you want to be more in control of the connection establishment for instance to add additional authentication mechanisms or maybe do some pooling with the connections.

As a reference here is an example of how a very similar API looks like in XAdES4j:
https://code.google.com/p/xades4j/wiki/Timestamping

Also iText has a similar interface for providing a custom implementation.

ebourg:
Ok, I didn't expect jsign to be used programmatically but that's nice to see you are able to push the limits. Let's merge this.

PR #20, opened 2016-03-23
MavenCentral publishing

nedtwigg:
I added the required sections to the pom.

One of the parts I struggled with when first publishing to mavencentral was GPG signing.  [I used bintray](https://bintray.com/docs/usermanual/uploads/uploads_syncingartifactswithmavencentral.html) to solve this issue.  The downside is that you have to create a bintray account (but they have GitHub OAuth), and you have to "request inclusion into jcenter" when you create a new project, but once you've done that it's fully automated from there on out.

I'm sure there's other ways to solve this problem too if you prefer :+1:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,17 @@
     </repository>
   </repositories>
 
+  <distributionManagement>
+    <snapshotRepository>
+      <id>ossrh</id>
+      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+    </snapshotRepository>
+    <repository>
+      <id>ossrh</id>
+      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+    </repository>
+  </distributionManagement>
+
   <dependencies>
     <dependency>
       <groupId>commons-cli</groupId>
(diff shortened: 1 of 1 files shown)


ebourg:
Thank you. We also have to declare the deployment repository, the nexus plugin and the gpg plugin:

http://central.sonatype.org/pages/apache-maven.html

nedtwigg:
I'm not a fan of maven, but I think I followed their tutorial correctly.  I added nexus-staging, gpg, and maven-source plugins.  I think you need to take the following actions to publish:

For deploying snapshots using nexus-staging, add this to your settings.xml:

```xml
<settings>
  <servers>
    <server>
      <id>ossrh</id>
      <username>your-jira-id</username>
      <password>your-jira-pwd</password>
    </server>
  </servers>
</settings>
```

For GPG, you need to have the `gpg` binary installed, and this in your settings.xml.
```xml
<settings>
  <profiles>
    <profile>
      <id>ossrh</id>
      <activation>
        <activeByDefault>true</activeByDefault>
      </activation>
      <properties>
        <gpg.executable>gpg2</gpg.executable>
        <gpg.passphrase>the_pass_phrase</gpg.passphrase>
      </properties>
    </profile>
  </profiles>
</settings>
```

If you want to set secure environment variables in travis and get the variables from there, you can use this syntax: `${env.VAR_NAME}`

ebourg:
Thank you for your help, jsign has now been published on Maven Central.

PR #24, opened 2016-10-02
Add PESignerBuilder

micpe083:
Added a builder for PESigner to share the build logic.

Please let me know your feedback.

Cheers,
Mike

### DIFF
--- a/src/main/java/net/jsign/PESigner.java
+++ b/src/main/java/net/jsign/PESigner.java
@@ -58,6 +58,7 @@ import net.jsign.asn1.authenticode.SpcIndirectDataContent;
 import net.jsign.asn1.authenticode.SpcPeImageData;
 import net.jsign.asn1.authenticode.SpcSpOpusInfo;
 import net.jsign.asn1.authenticode.SpcStatementType;
+import net.jsign.log.PELog;
 import net.jsign.pe.CertificateTableEntry;
 import net.jsign.pe.DataDirectoryType;
 import net.jsign.pe.PEFile;
@@ -75,6 +76,8 @@ import net.jsign.timestamp.TimestampingMode;
  * @since 1.0
  */
 public class PESigner {
+
+    private PELog log;
 
     private Certificate[] chain;
     private PrivateKey privateKey;
@@ -94,6 +97,14 @@ public class PESigner {
 
     public PESigner(KeyStore keystore, String alias, String password) throws NoSuchAlgorithmException, KeyStoreException, UnrecoverableKeyException {
         this(keystore.getCertificateChain(alias), (PrivateKey) keystore.getKey(alias, password.toCharArray()));
+    }
+
+    /**
+     * Set the logger.
+     */
+    public PESigner withLog(PELog log) {
+        this.log = log;
+        return this;
     }
 
     /**
@@ -199,7 +210,7 @@ public class PESigner {
         }
 
         try {
-            System.out.println("Adding Authenticode signature to " + file);
+            log.info("Adding Authenticode signature to " + file);
             sign(peFile);
         } catch (Exception e) {
             throw new SignerException("Couldn't sign " + file, e);
@@ -207,8 +218,7 @@ public class PESigner {
             try {
                 peFile.close();
             } catch (IOException e) {
-                System.err.println("Couldn't close " + file);
-                e.printStackTrace(System.err);
+                log.error("Couldn't close " + file, e);
             }
         }
     }
--- a/src/main/java/net/jsign/PESigner.java
+++ b/src/main/java/net/jsign/PESigner.java
@@ -16,6 +16,7 @@
 
 package net.jsign;
 
+import java.io.File;
 import java.io.IOException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
@@ -28,18 +29,6 @@ import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.List;
 
-import net.jsign.asn1.authenticode.AuthenticodeObjectIdentifiers;
-import net.jsign.asn1.authenticode.AuthenticodeSignedDataGenerator;
-import net.jsign.asn1.authenticode.SpcAttributeTypeAndOptionalValue;
-import net.jsign.asn1.authenticode.SpcIndirectDataContent;
-import net.jsign.asn1.authenticode.SpcPeImageData;
-import net.jsign.asn1.authenticode.SpcSpOpusInfo;
-import net.jsign.asn1.authenticode.SpcStatementType;
-import net.jsign.pe.CertificateTableEntry;
-import net.jsign.pe.DataDirectoryType;
-import net.jsign.pe.PEFile;
-import net.jsign.timestamp.Timestamper;
-import net.jsign.timestamp.TimestampingMode;
 import org.bouncycastle.asn1.ASN1Encodable;
 import org.bouncycastle.asn1.DERNull;
 import org.bouncycastle.asn1.DERSet;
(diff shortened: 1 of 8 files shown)


ebourg:
Thank you for the contribution Michael, this will be useful for the Maven plugin too. Before merging could you just adjust the following points please:
- make PESignerBuilder package private (it's for internal use only)
- adjust PESignerBuilder code style (opening curly braces at the end of the lines)
- PESignerBuilder is missing a license header
- no need to put final everywhere
- PESignerBuilder.param() should probably throw an IllegalArgumentException for unknown parameters

There is also an issue with message logging. With the Ant task we use the Ant logger, with the CLI we use System.err/out, but in PESignerBuilder only System.err/out is used. So we need a thin logging interface to redirect the messages properly. We could reuse the one from jdeb:

https://github.com/tcurdt/jdeb/blob/master/src/main/java/org/vafer/jdeb/Console.java
https://github.com/tcurdt/jdeb/blob/master/src/main/java/org/vafer/jdeb/ant/TaskConsole.java
https://github.com/tcurdt/jdeb/blob/master/src/main/java/org/vafer/jdeb/maven/MojoConsole.java

micpe083:
Hi Emmanuel, please see my feedback below:

- make PESignerBuilder package private (it's for internal use only)
 I think the builder should be public as the class it's building is also public. This will also give easy API access if someone needs it.

- adjust PESignerBuilder code style (opening curly braces at the end of the lines)
Done

- PESignerBuilder is missing a license header
Done

- no need to put final everywhere
Done

- PESignerBuilder.param() should probably throw an IllegalArgumentException for unknown parameters
Done

- Logger
Minimal logger added. Might be better to add slf4j or other api in future...

Cheers,
Mike

ebourg:
Thank you for the modifications. Regarding the visibility of PESignerBuilder, the PESigner class already provides a builder like API with a fluent syntax and strongly typed methods. From a user point a view I think it's preferable to use this class. PESignerBuilder is useful to factorize the codes in the CLI/Ant task/Maven plugin, but I don't think it'll have other use cases.

micpe083:
Looks like I forgot to mention this, but the reason I'm sending you this pull request is so I can use jsign without having to copy/paste the code in your cli/ant classes.

If I make the class private, then I can't use it and I'm back to square one again... :(

ebourg:
Do you intend to use PESignerBuilder outside of PESignerTask and PESignerCLI?

micpe083:
Yes

ebourg:
Could you give me an example of how you'll use it and why it's better than using PESigner directly?

micpe083:
I would use it like Option A below using the PEBuilder. Using PESigner directly (Option B) requires writing and maintaining much more code.

**Option A - create PESigner with PEBuilder:**

~~~~
PESignerBuilder builder = new PESignerBuilder(new PELogSysout(this));
builder.name(name);
builder.url(url);
builder.alg(algorithm);
builder.keystore(keystore);
builder.storepass(storepass);
builder.storetype(storetype);
builder.alias(alias);
builder.certfile(certfile);
builder.keyfile(keyfile);
builder.keypass(keypass);
builder.tsaurl(tsaurl);
builder.tsmode(tsmode);

PESigner signer = builder.build();
~~~~

**Option B - create PESigner without PEBuilder:**

~~~~
if (keystore != null && storetype == null) {
	// guess the type of the keystore from the extension of the file
	String filename = keystore.getName().toLowerCase();
	if (filename.endsWith(".p12") || filename.endsWith(".pfx")) {
		storetype = "PKCS12";
	} else {
		storetype = "JKS";
	}
}

PrivateKey privateKey;
Certificate[] chain;

// some exciting parameter validation...
if (keystore == null && keyfile == null && certfile == null) {
	throw new SignerException("keystore option, or keyfile and certfile options must be set");
}
if (keystore != null && (keyfile != null || certfile != null)) {
	throw new SignerException("keystore option can't be mixed with keyfile or certfile");
}

if (keystore != null) {
	// JKS or PKCS12 keystore
	KeyStore ks;
	try {
		ks = KeyStore.getInstance(storetype);
	} catch (KeyStoreException e) {
		throw new SignerException("keystore type '" + storetype + "' is not supported", e);
	}

	if (!keystore.exists()) {
		throw new SignerException("The keystore " + keystore + " couldn't be found");
	}
	FileInputStream in = null;
	try {
		in = new FileInputStream(keystore);
		ks.load(in, storepass != null ? storepass.toCharArray() : null);
	} catch (Exception e) {
		throw new SignerException("Unable to load the keystore " + keystore, e);
	} finally {
		try {
			if (in != null) {
				in.close();
			}
		} catch (IOException e) {
			// ignore
		}
	}

	if (alias == null) {
		throw new SignerException("alias option must be set");
	}

	try {
		chain = ks.getCertificateChain(alias);
	} catch (KeyStoreException e) {
		throw new SignerException(e.getMessage(), e);
	}
	if (chain == null) {
		throw new SignerException("No certificate found under the alias '" + alias + "' in the keystore " + keystore);
	}

	char[] password = keypass != null ? keypass.toCharArray() : storepass.toCharArray();

	try {
		privateKey = (PrivateKey) ks.getKey(alias, password);
	} catch (Exception e) {
		throw new SignerException("Failed to retrieve the private key from the keystore", e);
	}

} else {
	// separate private key and certificate files (PVK/SPC)
	if (keyfile == null) {
		throw new SignerException("keyfile option must be set");
	}
	if (!keyfile.exists()) {
		throw new SignerException("The keyfile " + keyfile + " couldn't be found");
	}
	if (certfile == null) {
		throw new SignerException("certfile option must be set");
	}
	if (!certfile.exists()) {
		throw new SignerException("The certfile " + certfile + " couldn't be found");
	}

	// load the certificate chain
	try {
		chain = loadCertificateChain(certfile);
	} catch (Exception e) {
		throw new SignerException("Failed to load the certificate from " + certfile, e);
	}

	// load the private key
	try {
		privateKey = PVK.parse(keyfile, keypass);
	} catch (Exception e) {
		throw new SignerException("Failed to load the private key from " + keyfile, e);
	}
}

if (alg != null && DigestAlgorithm.of(alg) == null) {
	throw new SignerException("The digest algorithm " + alg + " is not supported");
}

try {
	initializeProxy(proxyUrl, proxyUser, proxyPass);
} catch (Exception e) {
	throw new SignerException("Couldn't initialize proxy ", e);
}

PESigner signer = new PESigner(chain, privateKey)
		.withLog(log)
		.withProgramName(name)
		.withProgramURL(url)
		.withDigestAlgorithm(DigestAlgorithm.of(alg))
		.withTimestamping(tsaurl != null || tsmode != null)
		.withTimestampingMode(tsmode != null ? TimestampingMode.of(tsmode) : TimestampingMode.AUTHENTICODE)
		.withTimestampingAutority(tsaurl);

@SuppressWarnings("unchecked")
private Certificate[] loadCertificateChain(File file) throws IOException, CertificateException {
	FileInputStream in = null;
	try {
		in = new FileInputStream(file);
		CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
		Collection<Certificate> certificates = (Collection<Certificate>) certificateFactory.generateCertificates(in);
		return certificates.toArray(new Certificate[certificates.size()]);
	} finally {
		try {
			if (in != null) {
				in.close();
			}
		} catch (IOException e) {
			// ignore
		}
	}
}

private void initializeProxy(String proxyUrl, final String proxyUser, final String proxyPassword) throws MalformedURLException {
	// Do nothing if there is no proxy url.
	if (proxyUrl != null && proxyUrl.trim().length() > 0) {
		if (!proxyUrl.trim().startsWith("http")) {
			proxyUrl = "http://" + proxyUrl.trim();
		}
		final URL url = new URL(proxyUrl);
		final int port = url.getPort() < 0 ? 80 : url.getPort();

		ProxySelector.setDefault(new ProxySelector() {
			private List<Proxy> proxies = Collections.singletonList(new Proxy(Proxy.Type.HTTP, new InetSocketAddress(url.getHost(), port)));

			public List<Proxy> select(URI uri) {
				return proxies;
			}

			public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
			}
		});

		if (proxyUser != null && proxyUser.length() > 0 && proxyPassword != null) {
			Authenticator.setDefault(new Authenticator() {
				protected PasswordAuthentication getPasswordAuthentication() {
					return new PasswordAuthentication(proxyUser, proxyPassword.toCharArray());
				}
			});
		}
	}
}
~~~~

ebourg:
Ok understood, thank you for the example. Maybe more validation logic should be moved into PESigner, I'll think about it.

src/main/java/net/jsign/PESigner.java
```diff
@@ -163,17 +176,56 @@ public void sign(PEFile file) throws Exception {
...
+
         // compute the signature
         CertificateTableEntry entry = createCertificateTableEntry(file);
-        
+
         file.writeDataDirectory(DataDirectoryType.CERTIFICATE_TABLE, entry.toBytes());
         file.close();
     }
 
+    public void sign(String file) throws SignerException {
```
netmackan: This is just a shortcut for new File(String) but makes the API more unclear as PESigner.sign(String) does not tell the user that the String should be a file name. Better only have the strongly typed PESigner.sign(File).

netmackan:
I will continue to use the PESigner API which suites me very well and as I normally do not load the KeyStore instances from files and use custom Timestamper implementations and so on. That's why I prefer not mixing in to much high level logic in the PESigner class. That being said, having a simpler builder class in addition to that, as suggested here, could probably be useful to some.

ebourg:
I rebased the PR and made some additional modifications, thank you Michael.

PR #26, opened 2016-11-25
Add support for specifying signature algorithm and signature provider.

netmackan:
Before this patch the signature algorithm was assumed to be the digest algorithm with the key algorithm. While this works for simple cases it does not provide support when there are multiple signature algorithms available for a certain key type. For example signature algorithms using RSA keys are both RSASSA-PKCS1_v1.5 (SHAxWithRSA) and RSASSA-PSS (SHAxWithRSAandMGF1).
In addition it is maybe not necessarily so that the same hash algorithm must be used for both the digest in the PE, in the time-stamp token and as the digest of the signature.

This patch adds support for explicitly specifying which signature algorithm to use independent of what digest algorithm is used. It also adds support for specifying which provider that should be used for the signature. This is needed if the provider that should be used is not installed or if there are multiple providers available and one don't want to use the first one in the list.

### DIFF
--- a/src/main/java/net/jsign/PESigner.java
+++ b/src/main/java/net/jsign/PESigner.java
@@ -21,6 +21,7 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
+import java.security.Provider;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
@@ -76,7 +77,10 @@ public class PESigner {
 
     private Certificate[] chain;
     private PrivateKey privateKey;
-    private DigestAlgorithm algo = DigestAlgorithm.getDefault();
+    private DigestAlgorithm digestAlgorithm = DigestAlgorithm.getDefault();
+    private String signatureAlgorithm;
+    private Provider signatureProvider;
+    private String signatureProviderString;
     private String programName;
     private String programURL;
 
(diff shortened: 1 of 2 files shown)


ebourg:
Hi Markus, thank you for the patch. I wonder if specifying the provider is really necessary. Maybe the code could simply lookup the specified algorithm in the JDK and BouncyCastle providers, that should cover most of the cases.

netmackan:
I think you might be right that explicitly setting the provider normally shouldn't be needed. But for the case where the provider is not already installed one would need to first install it:
`Security.addProvider(new MyProvider());`
But thinking of it that should probably not be a problem. Do you want me to remove the setProvider() methods?

ebourg:
We can keep the provider, but I suggest using a unique method setting the algorithm and optionally the provider:

    withSignatureAlgorithm(String signatureAlgorithm)
    withSignatureAlgorithm(String signatureAlgorithm, Provider provider)

And in the sign method, if no provider was specified, we first check if the Java provider supports the specified algorithm, and if it doesn't we try the BouncyCastle provider.

netmackan:
Sure we can get rid of the setProvider methods and instead pair it with the withSignatureAlgorithm. That would just make it slightly less flexible for those that are fine with not specifying algorithm but need to specify provider. Those would now have to also specify signature algorithm. For my use case it doesn't matter though and there hasn't been support for specifying provider before anyway.

An other details it that Java/BouncyCastle provides both setProvider(Provider) and setProvider(String) and usages of those are maybe split 50-50 (just guessing) so some would want to just specify "BC" or "SunPKCS11-Slot2" or whatever their installed provider is called and some might not have an installed provider at all and would pass an Provider instance instead. To support all three cases it would be:

```
 withSignatureAlgorithm(String signatureAlgorithm)
 withSignatureAlgorithm(String signatureAlgorithm, Provider provider)
 withSignatureAlgorithm(String signatureAlgorithm, String provider)
```
Or to be completely flexible use the withSignatureProvider(Provider) and withSignatureProvider(String) as I suggested first.

> And in the sign method, if no provider was specified, we first check if the Java provider supports the specified algorithm, and if it doesn't we try the Bouncy Castle provider.

I am not sure I understand/know how one would do that?
As I understood it now when setProvider() is not called Java would take care of loading an Signature implementation from an installed Provider that supports the algorithm *and which supports the key*. So it could very well be that all though Bouncy Castle supports the algorithm if I am passing in a key instance created by the SunPKCS11 provider, i.e. an P11PrivateKey, BC would not be able to use it.

netmackan:
As we are a bit unsure about what will be best, I suggest we skip the provider part for now. In any case we will need withSignatureAlgorithm(String), so if we add that now we can always later add the provider methods.

What do you say should I remove the provider parts from this pull request so that it can be merged and open an other ticket for specifying the provider in the future?

ebourg:
Let's add the provider to the signature of the `withSignatureAlgorithm` method as you suggested (with the String variant). If someone complains we'll add the separate method setting the provider.

netmackan:
Okey, so only this two methods then?
```
 withSignatureAlgorithm(String signatureAlgorithm)
 withSignatureAlgorithm(String signatureAlgorithm, String provider)
```

ebourg:
and the third one with the Provider.

netmackan:
PR updated.

ebourg:
Merged, thank you!

PR #27, opened 2016-12-01
Fix HTTP Accept header to use right content-type for the response

netmackan:
Noticed that the TSP client code sets the MIME type for the request also in the "Accept" header but it should instead contain the MIME type for the expected response.

The response MIME type should be "application/timestamp-reply" according to RFC#3161 and as clarified in its errata.


ebourg:
Good catch, thank you!

Issue #25, opened 2016-11-25
Split Jsign into different sub modules


netmackan:
As discussed in PR #14 it would be good for projects reusing Jsign if it was split into different modules such as for instance:
- jsign-core
- jsign-cli
- jsign-ant
- jsign-maven-plugin

Then only the CLI and Ant jars would include the dependencies and jsign-core would be the right dependency for reusing jsign into other projects.

ebourg:
Done

PR #28, opened 2016-12-26
Suggestion for breaking up into modules

netmackan:
This pull requests tries to illustrate how a split of Jsign (#25) could look like.
I split out a jsign-core module and a second module jsign-app containing both the CLI and Ant Task. The jsign-app could potentially be split further if needed.

What do you think?

Note: This is not ready to be merged but for discussion. I made a mistake and deleted the site Git submodule and hasn't yet figured out how to revert or re-add it again...


ebourg:
Thank you for the suggestion Markus. The code is now split into a core, ant, cli and all modules. The all in one module retains the original artifactId (net.jsign:jsign) and still shades the dependencies. The core module doesn't shade any dependency and should be suitable for integrating jsign into your application.

PR #32, opened 2017-01-20
Correct a file descriptor leak in PEFile

blauwers:
This corrects a problem encountered when the PEFile tries to work on an invalid PEFile causing the constructor to fail without closing the `raf` handle. Otherwise, the system can run out of file descriptors if this error happens enough times.


ebourg:
Hi Bart,

Thank you for reporting this issue. With this modification the code doesn't compile because the final variables are set to null. Also, could you avoid reformatting the code (spaces were converted to tabs).

blauwers:
Hi Emmanuel,

I apologize for my sloppiness in the prior commit and have addressed the issues.

ebourg:
Merged, thank you!

PR #38, opened 2017-06-14
Allow alternative certificate chain with PKCS11

laurentgo:
Some PKCS11 devices (at least yubico) don't allow for the
whole certificate chain to be stored inside the token.

Allow jsign to be provided with an alternative certificate chain
so that the generated signature is considered valid.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/PESignerHelper.java
+++ b/jsign-core/src/main/java/net/jsign/PESignerHelper.java
@@ -40,10 +40,9 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 
-import sun.security.pkcs11.SunPKCS11;
-
 import net.jsign.pe.PEFile;
 import net.jsign.timestamp.TimestampingMode;
+import sun.security.pkcs11.SunPKCS11;
 
 /**
  * Helper class to create PESigner instances with untyped parameters.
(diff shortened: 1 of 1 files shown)


ebourg:
Thank you for the patch Laurent. This change was also suggested by @erwint in #3, but before merging it I think we have to consider the case where the token holds only the final certificate, and the certfile the intermediate certificates. In this situation the certificates must be merged to build the full chain.

laurentgo:
In my case, the certificate received from the CA was the full chain, which seemed pretty standard to me (Note that I received the certicate in both PEM and P7/SPC format, but for P7, the order of the list was backward, and that resulted in only the CA certificate to be added to the file to be signed).

Do you already have something in mind on how to solve this?

ebourg:
I'm thinking about a method appending a certificate to a chain, something like this:

    Certificate[] append(Certificate[] chain, Certificate cert)

with some sanity checks (don't add the certificate if it's already in the chain, ensure the issuer of the certificate is in the chain, and maybe sort the certificates if necessary).

On a side note, I'd be interested to know if jsign works well with the various certificates you got (PEM and PKCS7), because the unit tests cover a simple case with only one certificate in the chain for now.

laurentgo:
It worked well with the PEM certificate I received, but didn't with the P7 one (ordering issue).

ebourg:
Do you see the same behavior with the PEM file if you reverse the order of the certificates?

laurentgo:
Probably because I narrowed it down to that during my debugging session, but I will confirm it later when I have access back to my token.

Issue #40, opened 2017-09-04
Bug - Time service authority url at index 0 gets skipped


AtzDyna:
In Timestamper.timestamp in line 108 the variable count is incremented.  This means when line 110 is executed for the first time the variable count is already 1. Because of this, the first url retrieved from tsaurls is the url at index 1. The url at index 0 is skipped.

PR #41, opened 2017-09-04
Fixing bug (see issue #40)

AtzDyna:


ebourg:
Thanks a lot for the fix, I'll get that into the next release.

PR #43, opened 2017-10-13
withTimestampingAuthority() typo in PESigner 

aeons:


ebourg:
Good catch, thank you! This change will break the binary compatibility though. Could you add two deprecated methods with the old name to preserve the compatibility please?

aeons:
My editor fixed some whitespace issues as well I see :)

I haven't actually done Java for ages, so I'm not sure this even compiles.

ebourg:
Urgh white space changes! At least you squashed some tabulations ;)

PR #44, opened 2018-04-18
Add Inno Setup and install4j to the list of example EXE producers and fix link for exe4j

Vampire:


ebourg:
Thank you, I'm not sure it's desirable to list all existing executable generators though, 4 is good enough to illustrate the purpose of Jsign.

ebourg:
I agree that JSmooth looks abandoned and is probably worth removing from the list now.

Jsign is multi platform and mainly targeted at signing on non-Windows platforms where signtool isn't available. For this reason I focused the list on platform independent tools like NSIS and launch4j. Inno Setup runs only on Windows (or with Wine [too](https://katastrophos.net/andre/blog/2009/03/16/setting-up-the-inno-setup-compiler-on-debian/) with some tinkering), if it becomes some day available as a Debian package I'm willing to add it to the list.

PR #48, opened 2018-09-07
Fixed SunPKCS11 Provider instantiation with config file outside current directory.

konrade:
This patch fixes a problem when wanting to use a PKCS11 Provider with config residing outside current directory.
Seems there was some residue bug from commit 5ee0000cddc8c73504e2530c66679412d5410925


ebourg:
Merged, thank you!

Issue #55, opened 2019-05-29
Signature not recognized when using SHA-512


netmackan:
Way to reproduce:
1. Sign EXE with digest algorithm SHA-512 and no program name or program URL
2. Check digital signature details in Windows (7 for example)
Expected: Signature verified correctly.
Actual: "No signature was present in the subject." error message when looking at the details for the signature in the list.

Note that redoing the same steps but with SHA-256 or SHA-384 the signature is accepted.
Edit: Initially I thought the issue was missing program name and/or URL but it later turned out to not be the only issue.

ebourg:
For the record, not including the SPC_SP_OPUS_INFO structure has no impact on the validation of the SHA-512 signatures (tested on Windows 10). But for some obscure reason the `<<<Obsolete>>>` string in `SpcLink` derails the validation of the SHA-512 signatures only.

ebourg:
I've just figured out why the SHA-512 signatures are invalid with the `<<<Obsolete>>>` string, more details in #66.

PR #56, opened 2019-05-31
Make signatures with SHA-512 verify in Windows

netmackan:
Make output more similar to that from SignTool by:
- Filtering out CMS attributes
- Always including SPC SP OPUS Info
- Not using "<<<Obsolete>>>" string

With this changes SHA-512 is also verified correctly (issue #55).


jsign-core/src/main/java/net/jsign/asn1/authenticode/SpcLink.java
```diff
@@ -54,7 +54,7 @@ public ASN1Primitive toASN1Primitive() {
         } else if (moniker != null) {
             return new DERTaggedObject(false, 1, moniker);
         } else {
-            return new DERTaggedObject(false, 2, file);
+            return new DERTaggedObject(false, 2, new SpcString(""));
```
ebourg: What about changing the `file` field instead?

ebourg:
Thank you for the fix Markus. Is the removal of the signing time and CMS algorithm protection attributes required to validate the SHA-512 signature?

jsign-core/src/main/java/net/jsign/asn1/authenticode/SpcLink.java
netmackan: That would be better yes. Will update.

netmackan:
It turned out that not including the extra CMS attribute was not needed. The issue seems to have been only with the spc info.

jsign-core/src/main/java/net/jsign/asn1/authenticode/SpcLink.java
netmackan: Updated in one of the last commits.

ebourg:
Merged, thank you!

Issue #42, opened 2017-09-22
Signature verification only with Java.


denv77:
Hi. 
Please, help.
How can i verify signature only with Java?

I try this:

        BouncyCastleProvider prov = new BouncyCastleProvider();
        File file = new File("C:\\Windows\\SysWOW64\\jcPKCS11-2.dll");
        PEFile pef = new PEFile(file);
        List<CMSSignedData> signedDataList = pef.getSignatures();
        CMSSignedData cms = signedDataList.get(0);
        Store store = cms.getCertificates();
        SignerInformationStore signers = cms.getSignerInfos();
        Collection c = signers.getSigners();
        Iterator it = c.iterator();
        while (it.hasNext()) {
            SignerInformation signer = (SignerInformation) it.next();
            Collection certCollection = store.getMatches(signer.getSID());
            Iterator certIt = certCollection.iterator();
            X509CertificateHolder certHolder = (X509CertificateHolder) certIt.next();
            X509Certificate cert = new JcaX509CertificateConverter().setProvider(prov).getCertificate(certHolder);
            SignerInformationVerifier siv = new JcaSimpleSignerInfoVerifierBuilder().setProvider(prov).build(cert);
            System.out.println(signer.verify(siv));
        }

but I get an error:

Exception in thread "main" java.lang.NullPointerException
	at org.bouncycastle.cms.CMSSignedData$1.write(Unknown Source)
	at org.bouncycastle.cms.SignerInformation.doVerify(Unknown Source)
	at org.bouncycastle.cms.SignerInformation.verify(Unknown Source)
	at ru.centerinform.crypto.Main.main(Main.java:86)

ebourg:
I don't know sorry, but if you find out I can add a verify feature to jsign.

denv77:
Thanks for the answer. I'll try to find out.

denv77:
Hi.
Tell me please, why SHA1 digests do not equals?

I do this:

        BouncyCastleProvider prov = new BouncyCastleProvider();

        File file = new File("D:\\WORK\\PE Signature\\dll\\jcPKCS11-2-x32.dll");
        PEFile pef = new PEFile(file);
        byte[] psha1 = pef.computeDigest(DigestAlgorithm.SHA1);
        System.out.println("pef " + Arrays.toString(psha1));

        List<CMSSignedData> signedDataList = pef.getSignatures();
        CMSSignedData cms = signedDataList.get(0);
        Store store = cms.getCertificates();
[...]

PR #59, opened 2019-07-05
Verify method implemented

aravindarc:
A verification method has been implemented, which is in the TODO list.

After I couldn't find the cause of the exception: 
**Exception in thread "main" java.lang.NullPointerException 
at org.bouncycastle.cms.CMSSignedData$1.write(Unknown Source)** 
pointed by issue #42, implemented this verify method.

This is a shallow verify, if the exe file contains a chain of certificates, only the leaf certificates and their counter-signatures are verified. 

Check [this](https://www.ietf.org/rfc/rfc2315.txt) to understand the implementation.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/pe/PEFile.java
+++ b/jsign-core/src/main/java/net/jsign/pe/PEFile.java
@@ -16,34 +16,34 @@
 
 package net.jsign.pe;
 
-import java.io.Closeable;
-import java.io.File;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.PrintWriter;
+import java.io.*;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.channels.SeekableByteChannel;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
-import java.security.MessageDigest;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
+import java.security.*;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.*;
 
 import net.jsign.DigestAlgorithm;
 import net.jsign.asn1.authenticode.AuthenticodeObjectIdentifiers;
 
-import org.bouncycastle.asn1.ASN1Encodable;
-import org.bouncycastle.asn1.cms.Attribute;
-import org.bouncycastle.asn1.cms.AttributeTable;
-import org.bouncycastle.asn1.cms.ContentInfo;
+import org.bouncycastle.asn1.*;
+import org.bouncycastle.asn1.cms.*;
 import org.bouncycastle.asn1.pkcs.PKCSObjectIdentifiers;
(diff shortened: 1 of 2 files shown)


ebourg:
Thank you for the PR. I understand that this compares the digest of the file and the one in the signatures, but does it check the validity of the certificate chain?

aravindarc:
No it does not but if you could tell me if this is the way to approach verification, I'll try and add checking the validity of the certificate chain

ebourg:
@aravindarc The code has been refactored to support several file formats. Do you think you could rebase your PR on top of the latest changes? The `verify()` method could go in a new `AuthenticodeVerifier` class and take a Signable object as input.

aravindarc:
@ebourg Yes I'll try to change as you say and also validate the certificate chain.

mduft:
Hey. I copied the proposed code and use it to verify that our binaries are properly signed before releasing them. This works fine, the only thing I changed is to add a check for whether the used certificate is self-signed. I just copied the isSelfSigned() method from AuthenticodeSigner and verify the X509Certificate using this...

strager:
If this feature is added to Jsign, I can switch from osslsigncode to Jsign. Then, I can implement APPX signing (#81).

Is the only task for this PR resolving merge conflicts?

Context: I use osslsigncode for signing and verification of .exe files. I want to also sign APPX files, but this is not implemented in osslsigncode (or in Jsign). I'd rather implement APPX signing in Jsign than in osslsigncode, but that would only make sense if I use Jsign for signing and verifying .exe files. (I don't want to use both Jsign and osslsigncode.)

strager:
> If this feature is added to Jsign, I can switch from osslsigncode to Jsign. Then, I can implement APPX signing (#81).

I discovered [Relic](https://github.com/sassoftware/relic) which has all the features I need. Its interface isn't as nice as Jsign's for my use case, but I'll manage.

I no longer plan to work on #81.

ebourg:
> Is the only task for this PR resolving merge conflicts?

No I don't think I'll merge the PR, I'm working on another design that provides a detailed verification report.

kushalagrawal:
Hi,
I am looking for this windows executable implementation including the MSI file and I have been waiting for this pull request to be merged. I have tested this on a few executable (Exe and MSI) files. It mostly looks to be working for the MSI files as well. other than the part:
``` 
if(!Arrays.equals(computeDigest(signedDataAlgorithm), digestInfo.getDigest()))
                    throw new VerificationException("The embedded hash in the SignedData is not equal to the computed hash");
  ```
  
  Is there anything that could be done to validate the MSI files as well with this?

GitHubb3R:
> > Is the only task for this PR resolving merge conflicts?
> 
> No I don't think I'll merge the PR, I'm working on another design that provides a detailed verification report.

I'd like to ask about the status, progress of said that design, and whether its completely stalled.

ebourg:
Signature verification is now implemented, as an API and as a new command for the CLI.

https://ebourg.github.io/jsign/#verify-options

PR #65, opened 2019-11-11
Support signing of PowerShell scripts (Fix #64)

### LINKED ISSUE #64: Support for signing PowerShell scripts
You can sign PowerShell scripts by adding a signature block with the Authenticode signature.
It would be super nice if JSign would be able to do this too, so we can cross-platform sign our PowerShell scripts.

Vampire:
Closes #64


ebourg:
Nice, thank you! Introducing a base signer is a good idea, that will be necessary to support MSI files too. But I wonder if we could use a single CLI supporting different types of files, that would be simpler. The same goes for the Ant task and the Maven/Gradle plugins.

Vampire:
For sure, the "problem" is, that you called the things "pesign" or "signexe". It doesn't fit too well to sign PS or MSI with that.

ebourg:
You're right, it's time to rename them to "jsign" I guess.

Vampire:
Maybe, that would be a major breaking change though, so I didn't want to do that here.
And I'm going on vacation in two days, so will probably not have the time to do such a major change.
Besides that, you then also need to decide on how to differentiate which mode to use, whether with an additional paramter or something else and so on.
For now I also added some doc changes for PowerShell.

ebourg:
I've renamed pesign and signexe to jsign. I still have to refactor the code to support several file formats now.

ebourg:
I've completed the refactoring and rebased your implementation with a tweak to support multiple signatures per script. It's working great, thank you!

Vampire:
I only had a very brief look right now, but isn't there some facility missing to define the file type?
How do you sign a PS via Ant task, Maven goal or CLI?
You can use the API of course, but for the generic interfaces, shouldn't there be some "file type" option that might do some auto-detection based on content and / or file extension if it is not set explicitly?

ebourg:
The type of the file is detected automatically in `SignerHelper`. For PE files the header is inspected, and for PowerShell scripts it's guessed from the extension of the file.

Vampire:
Ah, I see.
But don't you think an explicit override from command-line would be sensible?
Sure, for EXE files, it should be properly parseable by header.
But relying on the file extension for PS is a bit restricting.
Think e. g. Linux, if someone writes some piece in PS and does it Linux-style, adding a shebang line that instructs the system to execute it as PS script and having no file extension for nicer usage.
To sign that with the current state you would need to rename the file just to sign it and then rename it back. An explicit option in the different UIs would better support this.

ebourg:
That sounds a bit far fetched, in a Linux context I don't think the Authenticode signature would be relevant. But I understand your point, and I agree that relying only on the file extension is a bit weak. Let's start simple and see how the feedbacks go, this can be easily changed later.

PR #67, opened 2019-12-02
Fix faulty merge of the PowerShell signing PR (#64, #65)

Vampire:
It is nice that you attributed the commit to me, despite the significant changes you had to do.
It is not so nice, that you introduced several major bugs that are now attributed to me too now. :-(

For example you cannot sign a PowerShell script through the CLI, because if you do not provide a script encoding it does not use UTF-8 as documented, but produces a `NullPointerException` and you cannot give an encoding explicitly either, as the parameter does not have an argument for the actual encoding anymore.

You also removed all doc-changes and changes to bash completion.
And you made it impossible to set the script encoding via API as you made the method package private.

But on my way to fix the faulty merge, I also caught two other bugs, so maybe it was worth the hassle. :-D
I hope I got all merge errors with this PR.

Btw. when do you plan to release v3, as we soon want / need to sign PowerShell scripts and I'd prefer using an official version. :-)

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -270,7 +270,8 @@ are provided to install it easily on most Linux distributions. On these systems 
 
 <pre>
   usage: jsign [OPTIONS] FILE
-  Sign and timestamp a Windows executable file.
+  Sign and timestamp a Windows executable file, a Microsoft Installer (MSI) or a
+  PowerShell script.
   
   -s,--keystore &lt;FILE>       The keystore file, or the SunPKCS11 configuration file
      --storepass &lt;PASSWORD>  The password to open the keystore
@@ -278,25 +279,29 @@ are provided to install it easily on most Linux distributions. On these systems 
                              - JKS: Java keystore (.jks files)
                              - PKCS12: Standard PKCS#12 keystore (.p12 or .pfx files)
                              - PKCS11: PKCS#11 hardware token
-  -a,--alias &lt;NAME>          The alias of the certificate used for signing in the keystore.
+  -a,--alias &lt;NAME>          The alias of the certificate used for signing in the
+                             keystore.
      --keypass &lt;PASSWORD>    The password of the private key. When using a keystore,
                              this parameter can be omitted if the keystore shares the
                              same password.
-     --keyfile &lt;FILE>        The file containing the private key (supports PEM &amp; PVK files)
+     --keyfile &lt;FILE>        The file containing the private key. PEM and PVK files
+                             are supported.
   -c,--certfile &lt;FILE>       The file containing the PKCS#7 certificate chain
                              (.p7b or .spc files).
   -d,--alg &lt;ALGORITHM>       The digest algorithm (SHA-1, SHA-256, SHA-384 or SHA-512)
   -t,--tsaurl &lt;URL>          The URL of the timestamping authority.
   -m,--tsmode &lt;MODE>         The timestamping mode (RFC3161 or Authenticode)
   -r,--tsretries &lt;NUMBER>    The number of retries for timestamping
-  -w,--tsretrywait &lt;SECONDS> The number of seconds to wait between timestamping retries
+  -w,--tsretrywait &lt;SECONDS> The number of seconds to wait between timestamping
+                             retries
   -n,--name &lt;NAME>           The name of the application
   -u,--url &lt;URL>             The URL of the application
      --proxyUrl &lt;URL>        The URL of the HTTP proxy
      --proxyUser &lt;NAME>      The user for the HTTP proxy. If an user is needed.
-     --proxyPass &lt;PASSWORD>  The password for the HTTP proxy user. If an user is needed.
+     --proxyPass &lt;PASSWORD>  The password for the HTTP proxy user. If an user is
+                             needed.
      --replace               Tells if previous signatures should be replaced.
-     --encoding &lt;ENCODING>   The encoding of the PowerShell script to be signed (UTF-8
+  -e,--encoding &lt;ENCODING>   The encoding of the PowerShell script to be signed (UTF-8
                              by default).
   -h,--help                  Print the help
 </pre>
--- a/jsign-core/src/main/java/net/jsign/SignerHelper.java
+++ b/jsign-core/src/main/java/net/jsign/SignerHelper.java
@@ -434,7 +434,7 @@ class SignerHelper {
         try (FileInputStream in = new FileInputStream(file)) {
             CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
             Collection<? extends Certificate> certificates = certificateFactory.generateCertificates(in);
-            return certificates.toArray(new Certificate[certificates.size()]);
+            return certificates.toArray(new Certificate[0]);
         }
     }
 
--- a/jsign-core/src/main/java/net/jsign/powershell/PowerShellScript.java
+++ b/jsign-core/src/main/java/net/jsign/powershell/PowerShellScript.java
@@ -57,6 +57,10 @@ public class PowerShellScript {
             "(?<signatureBlock>.*)" +
             "# SIG # End signature block\\r\\n");
 
+    private static final Pattern SIGNATURE_BLOCK_REMOVAL_PATTERN = Pattern.compile("(?s)" +
+            "\\r?\\n" +
+            "# SIG # Begin signature block.*$");
+
     private File file;
     private String content;
     private Charset encoding;
@@ -205,7 +209,7 @@ public class PowerShellScript {
      * @return the content without the signature
      */
     private String getContentWithoutSignatureBlock() {
-        return SIGNATURE_BLOCK_PATTERN.matcher(getContent()).replaceFirst("");
+        return SIGNATURE_BLOCK_REMOVAL_PATTERN.matcher(getContent()).replaceAll("");
     }
 
     public byte[] computeDigest(MessageDigest digest) {
--- a/jsign-core/src/test/java/net/jsign/PESignerTest.java
+++ b/jsign-core/src/test/java/net/jsign/PESignerTest.java
@@ -109,7 +109,7 @@ public class PESignerTest {
         try (FileInputStream in = new FileInputStream(new File("target/test-classes/keystores/jsign-test-certificate-full-chain.spc"))) {
             CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
             Collection<? extends Certificate> certificates = certificateFactory.generateCertificates(in);
-            chain = certificates.toArray(new Certificate[certificates.size()]);
+            chain = certificates.toArray(new Certificate[0]);
         }
         
         PrivateKey key = PrivateKeyUtils.load(new File("target/test-classes/keystores/privatekey-encrypted.pvk"), "password");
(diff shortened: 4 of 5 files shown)


Vampire:
I don't think the build errors are my fault, but an environment problem, for me the failing test runs fine here.

ebourg:
Thank you for the review and sorry for the bugs, I'll sort that out.

> you made it impossible to set the script encoding via API as you made the method package private.

This is intended, at the API level the encoding is defined in the `PowerShellScript` constructor.

> when do you plan to release v3, as we soon want / need to sign PowerShell scripts and I'd prefer using an official version. :-)

There is a big push coming soon with support for MSI signing and some other fixes. I'd like to release Jsign 3.0 before the end of the month.

ebourg:
> 
> 
> I don't think the build errors are my fault, but an environment problem, for me the failing test runs fine here.

Yes travis is broken, the timestamping requests keep timing out during the tests.

Vampire:
> Thank you for the review and sorry for the bugs, I'll sort that out.

To sort it out, just merge my PR. ;-)

> This is intended, at the API level the encoding is defined in the `PowerShellScript` constructor.

Ah, sorry, I also missed that the `sign(File)` variants are package private too.
I reverted that visibility change now.

> There is a big push coming soon with support for MSI signing and some other fixes. I'd like to release Jsign 3.0 before the end of the month.

Cool, thanks.

ebourg:
> To sort it out, just merge my PR. ;-)

I can't because there are too many unrelated changes in a single commit, but I'll extract and merge the relevant parts separately.

ebourg:
The NPE is now fixed but I still have to merge some of your changes (the CLI help message displayed when the file isn't specified, the PS pattern matching changes and the documentation update).

Vampire:
> I can't because there are too many unrelated changes in a single commit.

Sorry, I rebased and split into single commits now.

ebourg:
Thank you, it's easier to review.

ebourg:
I have a question regarding the encoding of the PowerShell scripts. signtool doesn't seem to have an option to set the encoding, and when Windows checks the signature (when inspecting the signatures in the file properties window) it doesn't ask for the encoding. So how can the signature/verification be right? Is the script converted to UTF-8 when signed? Is the encoding automatically detected?

ebourg:
> 27f0ff9 Add PowerShell scripts to the abstract about JSign and fix dead links…

I've fixed the exe4j link but left out the references to the other exe generators. There are already many examples and I'd rather not turn the summary into an exhaustive catalog of existing tools.

> 1e33c65 Add documentation about --scriptEncoding to index.html
> cec8d29 Add PowerShell scripts to the man page
> 82f56db Add --scriptEncoding to the man page
> a43a19f Update help-output in index.html file
> f966e41 Mention PowerShellSigner too where PESigner is mentioned on API usage…

I made similar documentation changes on my branch already.

> e8dd87c Replace long deprecated <<-syntax by doLast-syntax in Gradle example

Applied

> af30827 Do not fail with exception if signee is not given on CLI but show hel…

Instead of displaying the help I've opted for printing an explicit error message.

> 939ec62 Use empty array as toArray-argument as is recommended since Java 6

Applied

> b14d710 Throw a dedicated exception if the PowerShell script cannot be read f…

Applied

> 45bb8e8 Add a PowerShellScript constructor with content and encoding

I've removed the constructor taking a String so I didn't apply this change. I'm tempted to introduce a String constructor to create a PEFile/MSIFile/PowerShellScript from a filename, and this constructor would get in the way.

> d16792d Fix signing of PowerShell scripts in some cases

I've picked the regex with the named capturing group.

I've left out the `getContent()` -> `content` substitution because I think it's better to call the method if the class is ever derived.

I failed to understand the expression `"(?<=^|\n)# |\r\n"`, does this address the line ending issue you mentioned in the commit message? An example would be nice to write the corresponding test case.

> bf0d4e1 Remove the explicit programmatic default encoding in the Mojo 

I applied it but reverted it because it broke the unit tests. I'm too lazy to fix the tests :)


The changes are now on the master branch. Thanks again for the help.

Vampire:
> I have a question regarding the encoding of the PowerShell scripts. signtool doesn't seem to have an option to set the encoding, and when Windows checks the signature (when inspecting the signatures in the file properties window) it doesn't ask for the encoding. So how can the signature/verification be right? Is the script converted to UTF-8 when signed? Is the encoding automatically detected?

I'm not 100% sure, but I think the default windows encoding is used, except maybe if some BOM is found. Maybe there is also some detection, I don't know.
I think the same encoding that is used for running is used for signing and verifying signature.
As far as I know, the encoding depends on the default windows encoding which can be different on different machines.
Actually if some BOM is found, like in a UTF-16 file with BOM, the respective encoding is used to read and execute the script. You can easily try this by printing `Hello wörld` and checking whehter the `ö` is printed correctly.
Interestingly, if you try to sign the script with PowerShell using `Set-AuthenticodeSignature` and it is in UTF-16 with BOM, you only get an `UnknownError`.

So the best to write scripts that others should use, is to only use ASCII characters in the script, because those should be the same in most encodings.
I chose UTF-8 as default encoding, as hopefully the average script should only use ASCII characters and so `UTF-8` should work fine and if not, one can manually choose which encoding to use.

The alternative would be to reverse-engineer PowerShell to find out how the encoding is determined exactly and keep this in sync for the future too if there are any changes and I was not up to doing this.

One could of course also use the platform default encoding, but that means it behaves different on different machines and I really dislike this, so thought a defined default encoding is better in any way and UTF-8 is usually a good choice.

Kotlin for example followed the same road. Unlike Java which uses the platform default encoding when translating bytes to text, Kotlin uses UTF-8 if nothing else is specified explicitly which ususally makes more robust and expected code.

> > [82f56db](https://github.com/ebourg/jsign/commit/82f56dbbef1b2f145facf17b25afb2a11f089d94) Add --scriptEncoding to the man page
> > [a43a19f](https://github.com/ebourg/jsign/commit/a43a19f7b78538ff2cc77f873de42d8b0eeb10b4) Update help-output in index.html file
> 
> I made similar documentation changes on my branch already.

The help output in `index.html` is still not like the real output and in both cases the short option `-e` is missing.

> > [939ec62](https://github.com/ebourg/jsign/commit/939ec6267500a88c6a820e7c4647e71378f1b1db) Use empty array as toArray-argument as is recommended since Java 6
> 
> Applied

Not fully, you only applied one of four occurrences.

> > [d16792d](https://github.com/ebourg/jsign/commit/d16792d2d998e0f7cc64500d74aa7bbc4c9a505d) Fix signing of PowerShell scripts in some cases
> 
> I've picked the regex with the named capturing group.
> 
> I've left out the `getContent()` -> `content` substitution because I think it's better to call the method if the class is ever derived.

Makes sense, but you also left out the most important part which makes Jsign behave different from `Set-AuthenticodeSignature` and produce an invalid signature.

> I failed to understand the expression `"(?<=^|\n)# |\r\n"`, does this address the line ending issue you mentioned in the commit message? An example would be nice to write the corresponding test case.

The regex is basically a combination of the other two with additionally only replacing `# ` at the start of a line which makes the regex more efficient as it doesn't need to look for `# ` if it is not at the start of a line and only needs to do one replace run instead of two.

`(?<=^|\n)# |\r\n` means either `(?<=^|\n)# ` or `\r\n`
`(?<=^|\n)# ` means `# ` that is preceded by `^|\n` which means start of string or newline character
`\r\n` means CRLF

> > [bf0d4e1](https://github.com/ebourg/jsign/commit/bf0d4e19f7f228d604e141e41f4ee45024b9d56d) Remove the explicit programmatic default encoding in the Mojo
> 
> I applied it but reverted it because it broke the unit tests. I'm too lazy to fix the tests :)

Yeah, also seen the failing tests and also removed that commit from my PR already.
I thought the tests run for me, but maybe I was wrong.

---

I updated my PR with the still missing parts. :-)

ebourg:
I started experimenting with different encodings. At least an UTF-16LE file, without BOM, is properly handled by signtool. The signature is appended in UTF-16 and Windows manages to validate it. I'll try with other encodings, with and without BOM.

ebourg:
> 62c309b Update help-output in index.html file
> e7c0500 Use empty array as toArray-argument as is recommended since Java 6
> dd8fd50  Add -e short option for --encoding to the man page

Applied

> 638c20c Fix signing of PowerShell scripts in some cases

Sorry but I fail to understand the issue. If the line separator is changed, technically the content is different and the signature should become invalid. After all, detecting alterations is the purpose of signing. But even if we assume the signature should still be valid, I don't understand how this commit addresses the issue.

Vampire:
> Update help-output in index.html file

Not fully, the text in the `index.html` still does not match the actual output.
And I do not only mean the line breaks, but also the text, e. g. for the keyfile option. :-)

> Sorry but I fail to understand the issue. If the line separator is changed, technically the content is different and the signature should become invalid. After all, detecting alterations is the purpose of signing. But even if we assume the signature should still be valid, I don't understand how this commit addresses the issue.

Yes, indeed you didn't understand me correctly.
Just do as mentioned in the commit.
Have a signed script with valid signature.
Change the line endings to LF in the whole file.
Now the signature is invalid, or rather there is no signature anymore as the signature block seems to have to use CRLF.
Now try again to sign the file as-is.
This results in an invalid signature being added.
With my change, the signature added is valid and the wrong-line-ended signature block is removed which is exactly the result that the PowerShell cmdlet `Set-AuthenticodeSignature` produces.

ebourg:
Thank you for the clarification, I'm starting to understand. I didn't notice that the signature was always added with CRLF instead of reusing the file EOL. So if the file uses LF, once signed it gets a mix of LF for the content and CRLF for the signature. I've verified that:

1. if the file is converted to all CRLF, the signature is indeed invalidated.
2. if the file is converted to all LF, Windows no longer sees the signature, nor does the `Get-AuthenticodeSignature` cmdlet. But `Set-AuthenticodeSignature` manages to replace the signature. In this situation Jsign currently adds a second signature block.

I'll write a test covering this case.

ebourg:
I think this is fixed now, please give it a look.

> Not fully, the text in the index.html still does not match the actual output.
> And I do not only mean the line breaks, but also the text, e. g. for the keyfile option. :-)

I modified slightly the help to fit better on the web page, that's why it's different. The main description still needs to be updated but it will change again once VBScript/JScript support is implemented, so I leave that out for now.

Vampire:
> Thank you for the clarification, I'm starting to understand. I didn't notice that the signature was always added with CRLF instead of reusing the file EOL. So if the file uses LF, once signed it gets a mix of LF for the content and CRLF for the signature. I've verified that:

Exactly.

>     1. if the file is converted to all CRLF, the signature is indeed invalidated.
> 
>     2. if the file is converted to all LF, Windows no longer sees the signature, nor does the `Get-AuthenticodeSignature` cmdlet. But `Set-AuthenticodeSignature` manages to replace the signature. In this situation Jsign currently adds a second signature block.

Exactly, that was what I observed too and added this second regex for. :-)

> I'll write a test covering this case.

Thanks

> I think this is fixed now, please give it a look.

lgtm now

> I modified slightly the help to fit better on the web page, that's why it's different. The main description still needs to be updated but it will change again once VBScript/JScript support is implemented, so I leave that out for now.

Ah, ok, if it is intentionally it's fine of course. :-) 👍

PR #75, opened 2020-03-05
--alias option optional if keystore contains only one alias

mlocati:
If the keystore contains only one alias, what about making the `--alias` CLI option optional? We can retrieve it from the keystore itself...


ebourg:
Good idea, thank you for the patch. I'll merge it into the next release.

Issue #77, opened 2020-04-29
What is "checksum" vs. "checksum (computed)"?


netmackan:
Hi Emmanuel,

I noticed that Jsign does not update the checksum after the certificate table is written and found a possible bug in PEFile.updateChecksum() that it does not position the buffer to write. I can make a pull request for that but I first have a question so that I can check if it is working:

Question: When printing the info from a PE file it can be seen that checksum (i.e. the value from the checksum field) and the "checksum (computed)" is different. Are they not supposed to be the same or are them somehow in different format/encoding?

```
Checksum:                           0x234242b
Checksum (computed):        0x1234d
```

The reason I ask is that if I fix the potential issue with the checksum not being updated I want to know what value should be written and I do not understand why the  computed checksum is different (even if the file has not changed) and if I can use that...

Cheers,
Markus

ebourg:
The computed checksum is the expected value of the checksum field if it was correct. AFAIK the checksum field is never used to ensure the integrity of the executable, so its value doesn't really matter.

ebourg:
> However, it seems that some applications actually do check the checksum field and refuses to run if it is incorrect. I haven't fully confirmed this is the case yet but it would explain an issue that we see.

If there are applications checking this field then I agree it's important to ensure it's right.


> I was thinking that perhaps you know or don't know the status of the checksum calculation code? My thinking is that if it is not needed for the signature verification then maybe it is not tested as much and there could be some issue there?

There might be corner cases not properly handled, I haven't tested this sufficiently to garantee it's perfect.

netmackan:
I found that the checksum calculation only works if the file size is less than 64 KB. I will submit a pull request for fixing the checksum calculation and also so it works with larger files.

PR #78, opened 2020-05-19
Fix for checksum update and calculation

netmackan:
Two fixes as discussed in #77 :
* Position the buffer so that it will be written out to the checksum field. This solves the issue with the field not being updated.
* Clear the buffer so that it can be used for the next read. This solves the issue with incorrect checksum calculation for files larger than the buffer (64 KiB).


ebourg:
Thank you Markus, I'll get a look and add a unit test.

ebourg:
Patch applied, thank you!

PR #85, opened 2020-10-07
better proxy management to be able to execute tests behind a proxy

lionnel:
Hi,

I'have isolated the initializeProxy to be able to set environment http.proxyUrl, http.proxyUser & http.proxyPassword to be able to pass local proxy.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/SignerHelper.java
+++ b/jsign-core/src/main/java/net/jsign/SignerHelper.java
@@ -369,7 +369,7 @@ class SignerHelper {
         }
 
         try {
-            initializeProxy(proxyUrl, proxyUser, proxyPass);
+            initializeProxy(proxyUrl, proxyUser, proxyPass, console);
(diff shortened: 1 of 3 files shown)


ebourg:
Thank you for the suggestion, but is this really necessary? The JVM automatically picks the `http.proxy*` system properties to configure the proxy. We shouldn't have to redo the same work here.

lionnel:
Yes the JVM take account of proxy url (host and port).
But as far as I have checked it do not take account of the user/password part.
No pb if you do not merge the pr.
I'm trying to implement a verifier. As I work being an authenticate proxy, it's more to me to be sure not to push this part in the future pr.
Regards

ebourg:
The username/password properties should be used automatically, at least with Java 8:

https://stackoverflow.com/questions/1626549/authenticated-http-proxy-with-java

Also if you work behind a proxy, you may want to try `-Djava.net.useSystemProxies=true`.

Issue #91, opened 2021-06-30
Unable to sign binaries with GOOGLECLOUD


rvadim:
**STR:**
1.  Generate asymmetric key and CA
```
gcloud kms keyrings create test --location <region>```
gcloud kms keys create my-key --keyring test --location <region> --purpose "asymmetric-signing" --default-algorithm "rsa-sign-pkcs1-2048-sha256"
openssl genrsa -out ca.key 2048
```
2. Create CSR and Certificate
```
git clone https://github.com/mattes/google-cloud-kms-csr
cd google-cloud-kms-csr
go build -o csr
./csr -key $(gcloud kms keys versions list  --key my-key --keyring test --location=<region>) -out my.csr --common-name MyOrg
openssl -req -in my.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out my.crt
```
3. Run jsign
```
jsign --storetype GOOGLECLOUD
  --storepass "$(gcloud auth print-access-token)"
  --keystore "projects/<project>/locations/<location>/keyRings/test"
  --alias "my-key"
  --certfile my.crt
  my.exe
```

**Expected result:**
Binary signed

**Actual result:**
```
jsign: Couldn't sign my.exe
java.lang.RuntimeException: java.security.KeyStoreException: java.io.IOException: HTTP Error 404 - Not Found
	at net.jsign.jca.SigningServiceKeyStore.engineAliases(SigningServiceKeyStore.java:88)
	at java.base/java.security.KeyStore.aliases(KeyStore.java:1269)
	at net.jsign.SignerHelper.build(SignerHelper.java:338)
	at net.jsign.SignerHelper.sign(SignerHelper.java:476)
	at net.jsign.JsignCLI.execute(JsignCLI.java:113)
	at net.jsign.JsignCLI.main(JsignCLI.java:40)
Caused by: java.security.KeyStoreException: java.io.IOException: HTTP Error 404 - Not Found
	at net.jsign.jca.GoogleCloudSigningService.aliases(GoogleCloudSigningService.java:81)
	at net.jsign.jca.SigningServiceKeyStore.engineAliases(SigningServiceKeyStore.java:86)
	... 5 more
Caused by: java.io.IOException: HTTP Error 404 - Not Found
	at net.jsign.jca.RESTClient.query(RESTClient.java:76)
	at net.jsign.jca.RESTClient.get(RESTClient.java:43)
[...]

PR #92, opened 2021-06-30
Use key path instead of key name for google signer

rvadim:
Fixing https://github.com/ebourg/jsign/issues/91


jsign-core/src/main/java/net/jsign/SignerHelper.java
```diff
@@ -321,7 +321,7 @@ private AuthenticodeSigner build() throws SignerException {
             String[] elements = storepass.split("\\|");
             provider = new SigningServiceJcaProvider(new DigiCertOneSigningService(elements[0], new File(elements[1]), elements[2]));
         } else if ("GOOGLECLOUD".equals(storetype)) {
-            provider = new SigningServiceJcaProvider(new GoogleCloudSigningService(keystore.getName(), storepass, alias -> {
+            provider = new SigningServiceJcaProvider(new GoogleCloudSigningService(keystore.getPath(), storepass, alias -> {
```
rvadim, line 324: Looks like we will get error on Windows here. I will try to fix it.

ebourg:
Good catch, thank you!

Issue #95, opened 2021-07-13
Use RSA as digest encryption algorithm instead of sha256RSA


yaegor:
Currently, jsign-signed EXE files have "sha256RSA" (1.2.840.113549.1.1.11) as "Digest encryption algorythm" in the signature.
See ![screenshot](https://user-images.githubusercontent.com/3315/125463248-b2b12623-6064-4417-a27d-3001e40b4855.png).

Though this does not seem to affect Windows signature verification, it seems common to use "RSA" (1.2.840.113549.1.1.1) instead. Microsoft's signtool and osslsigncode produce signatures with RSA instead of sha256RSA.
Also, the only available(?) [Authenticode specification](http://download.microsoft.com/download/9/c/5/9c5b2167-8017-4bae-9fde-d599bac8184a/authenticode_pe.docx) (from 2008 and obsolete, but still) mentions only RSA and DSA as supported digestEncryptionAlgorithm.
It seems appropriate to change jsign behavior to produce signatures with "RSA" digest encryption algorythm or at least provide an option to set the algorythm.

ebourg:
This field is managed by Bouncy Castle. If there is no impact on the validity nor the size of the signature I'm tempted to leave it as is for now. I think the issue should be reported to Bouncy Castle first, the `DefaultCMSSignatureEncryptionAlgorithmFinder` class handles the algorithms up to `sha1withRSA`, I wonder if there is reason justifying the lack of recent algorithms or if this was simply overlooked.

yaegor:
Thank you for the reply.
[DefaultCMSSignatureEncryptionAlgorithmFinder](https://github.com/bcgit/bc-java/blob/b35d626619a564db860e59e1cda353dec8d7a4fa/pkix/src/main/java/org/bouncycastle/cms/DefaultCMSSignatureEncryptionAlgorithmFinder.java#L46) in Bouncy Castle references [RFC5754 section 3.2](https://datatracker.ietf.org/doc/html/rfc5754#section-3.2) which, if I am reading this correctly, says "old, sha1-times rsaEncryption can be used, but new shaXXXWithRSAEncryption can be used as well":
> [RFC3370], Section 3.2, specifies the conventions for RSA with SHA-1 (RSASSA-PKCS1-v1_5) public key algorithm identifiers, parameters, public keys, and signature values.
[...]

PR #96, opened 2021-07-13
Force digest encryption algorithm to "RSA" instead of "sha256RSA" (#95)

yaegor:
Simple solution forcing RSA for the specific case of "sha256RSA" algorythm (#95).
It might be appropriate to force RSA for all algorithms or add an option instead.


ebourg:
Thank you for the patch but this addresses only the `sha256RSA` case, I guess there is a similar issue with `sha384RSA` and `sha512RSA`. If I read correctly the `DefaultCMSSignatureEncryptionAlgorithmFinder` implementation the SHA-1 case is already handled.

Issue #82, opened 2020-07-22
Signing large MSI files fails


Sasham111:
Hi,
I tried to use Jsign to sign a large MSI file (311MB), it fails with the following Exception:

`[ec2-user@ip-172-31-35-244 original_jsign]$ jsign -s key.jks --storepass 123456 -storetype jks -a EKM ~/ekm-2.0_large.msi
Adding Authenticode signature to /home/ec2-user/ekm-2.0_large.msi
jsign: Couldn't sign /home/ec2-user/ekm-2.0_large.msi
java.lang.RuntimeException: java.io.IOException: Map failed
        at net.jsign.poi.poifs.filesystem.POIFSStream$StreamBlockByteBufferIterator.next(POIFSStream.java:170)
        at net.jsign.poi.poifs.filesystem.POIFSStream$StreamBlockByteBufferIterator.next(POIFSStream.java:142)
        at net.jsign.msi.MSIFile.computeDigest(MSIFile.java:180)
        at net.jsign.msi.MSIFile.createIndirectData(MSIFile.java:199)
        at net.jsign.AuthenticodeSigner.createSignedData(AuthenticodeSigner.java:368)
        at net.jsign.AuthenticodeSigner.sign(AuthenticodeSigner.java:339)
        at net.jsign.SignerHelper.sign(SignerHelper.java:424)
        at net.jsign.JsignCLI.execute(JsignCLI.java:111)
        at net.jsign.JsignCLI.main(JsignCLI.java:40)
Caused by: java.io.IOException: Map failed
        at sun.nio.ch.FileChannelImpl.map(FileChannelImpl.java:938)
        at net.jsign.poi.poifs.nio.FileBackedDataSource.read(FileBackedDataSource.java:95)
        at net.jsign.poi.poifs.filesystem.POIFSFileSystem.getBlockAt(POIFSFileSystem.java:427)
        at net.jsign.poi.poifs.filesystem.POIFSStream$StreamBlockByteBufferIterator.next(POIFSStream.java:166)
        ... 8 more
Caused by: java.lang.OutOfMemoryError: Map failed
        at sun.nio.ch.FileChannelImpl.map0(Native Method)
        at sun.nio.ch.FileChannelImpl.map(FileChannelImpl.java:935)
        ... 11 more
`

Signing a small file of 4MB completes successfully. Is there any size limitation of MSI files that can be signed? If so, what is it?
Thanks in advance

ebourg:
[...]

PR #99, opened 2021-08-12
Workaround for "OutOfMemoryError: Map failed" error when signing large MSI files

yaegor:
This fixes OOM for large MSI files for me (#82).
It is hacky, but I am not sure what is the proper API to iterate POIFSDocument.


ebourg:
Closing since this was solved with a different approach.

PR #126, opened 2022-05-05
Fix attaching signatures for PE files

markt-asf:
I spotted this testing repeatable builds with Apache Tomcat. This might not be the right way to fix the issue generally but it does fix the issue I was seeing. In short, when re-attaching the signature the PE file needs to be padded to a multiple of 8 bytes first - just like it is when signing the PE file directly.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/SignerHelper.java
+++ b/jsign-core/src/main/java/net/jsign/SignerHelper.java
@@ -547,6 +547,14 @@ class SignerHelper {
                 return;
             } catch (Exception e) {
                 throw new SignerException("Couldn't attach the signature to " + file, e);
+            } finally {
+                try {
+                    if (signable instanceof Closeable) {
+                        ((Closeable) signable).close();
+                    }
+                } catch (IOException ioe) {
+                    // Ignore for now
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


ebourg:
The non padded file was the NSIS installer?

markt-asf:
Hmm. My proposed patch may have triggered the file locking issue again.

ebourg:
I'm not even sure the checksum is properly computed when the file isn't padded, the tests don't cover this case.

I'd expected the code to:
1. compute the checksum as if the file was padded
2. pad the file only when the signature table is added, i.e. in `PEFile#setSignature()`, currently this is performed by `AuthenticodeSigner#sign()` and it doesn't feel at the right place, the signer code should be format agnostic.

I'll investigate further, thank you for reporting this.

ebourg:
I've made a couple of changes that should cover the case of a non padded file signed with a detached signature, could you give it a try? I haven't looked at the file locking issue yet.

markt-asf:
LGTM. I re-ran my signing the uninstaller test and adding the detached signature results in a valid signature (as far as Windows is concerned) whereas before your change it would result in an invalid signature.

ebourg:
Thank you for checking, I've also merged the locking fix. Time to release now.

ebourg:
> I did wonder if refactoring signer.sign() and attach() so they both accepted a Signable instance would help.

I've implemented your suggestion. `AuthenticodeSigner.sign()` no longer closes the file, `Signable` is now `Closeable`, attach/detach work directly on the signable, and a single try-with-resource encompasses the whole sign/attach/detach operation. That's much cleaner!

PR #128, opened 2022-06-02
Add ignorekeycert option

laurentgo:
Add an option to ignore certificate chain provided by the store and use
the one provided by the user without any check.

### DIFF
--- a/jsign-ant/src/main/java/net/jsign/JsignTask.java
+++ b/jsign-ant/src/main/java/net/jsign/JsignTask.java
@@ -87,6 +87,9 @@ public class JsignTask extends Task {
     /** Tells if a detached signature should be generated or reused. */
     private boolean detached;
 
+    /** Tells if the certificate provided by the keystore should be ignored. */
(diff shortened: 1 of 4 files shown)


ebourg:
I'm not sure to understand the use case, in what situation one would want to replace the certificate from the keystore? Why not updating the certificate on the keystore instead? I don't mind replacing the certificate but I would prefer something easier to use, for example if the certfile specified contains the signing certificate, use it automatically instead of the one from the keystore.

laurentgo:
Sorry, it has been a while since I opened the issue but at the time I opened the issue and proposed the change, it was to support some remote keystore system which would store securely a private/public key but could not associate a certificate chain with it. The PKCS11 layer would just generate a fake certificate because it is kind of mandated by the API, but I would need to pass the actual certificate generated with the key and signed by the actual CA to esign

ebourg:
The certificate specified with the certfile parameter now overrides the keystore certificate by default.

PR #129, opened 2022-06-02
Compare public keys instead of the full certificate

laurentgo:
Only compare public keys between the keystore certificate and the
certificate chain provided by the user instead of the full certificate.


ebourg:
The code has been updated on the master branch,  it no longer checks if the certificate from the specified certfile parameter matches the certificate from the keystore. So you should be able to specify a different certificate as long as the keys match. If the keys don't match a verification error is thrown after signing the file.

Issue #139, opened 2022-12-17
DigestAlgorithmIdentifiers now omits null parameters which causes issues on some systems


netmackan:
It has been reported that the signatures changed between one version and an other and we have tracked this down to be a change introduced in Bouncy Castle version 1.69 (probably) where it stopped including the NULL parameters in SHA2 DigestAlgorithmIdentifier:s in CMSSignedData.

As far as I can tell this change follows [RFC 5754](https://www.rfc-editor.org/rfc/rfc5754#section-2) and should be the right thing for CMS in general but I am not sure if it is right for Authenticode (?). At least we have got reports that systems are having problem with this change and also this is a difference as compared to what signtool does which still generates the NULL parameters.

Diff between Jsign and SignTool shows that SignTool still includes this NULL parameters:
```
--- HelloPE-signtool.exe.peprint	2022-12-16 11:18:33.805000000 +0100
+++ HelloPE-dss.exe.peprint	2022-12-16 11:40:11.468000000 +0100 
 PE Header
   Machine:                    I386
@@ -30,8 +30,8 @@
   Subsystem version:          4.0
   Size of image:              32768
   Size of headers:            512
-  Checksum:                   0x114c6
-  Checksum (computed):        0x114c6
+  Checksum:                   0xe043
+  Checksum (computed):        0xe043
   Subsystem:                  WINDOWS_CUI
   DLL characteristics:        0x0
   Size of stack reserve:      1048576
@@ -43,7 +43,7 @@
 Data Directory
   IMPORT_TABLE                   0x00002018       79 bytes
   RESOURCE_TABLE                 0x00004000      736 bytes
-  CERTIFICATE_TABLE              0x00000c00     2760 bytes
+  CERTIFICATE_TABLE              0x00000c00     2800 bytes
   BASE_RELOCATION_TABLE          0x00006000       12 bytes
   IMPORT_ADDRESS_TABLE           0x00002010        8 bytes
   CLR_RUNTIME_HEADER             0x00002064       72 bytes
@@ -66,7 +66,6 @@
             DER Set
                 DER Sequence
                     ObjectIdentifier(2.16.840.1.101.3.4.2.1)
-                    NULL
             DER Sequence
                 ObjectIdentifier(1.3.6.1.4.1.311.2.1.4)
                 Tagged [0]
@@ -305,7 +304,6 @@
                         Integer(7516069195698284833)
                     DER Sequence
                         ObjectIdentifier(2.16.840.1.101.3.4.2.1)
[...]

PR #140, opened 2022-12-18
Make sure digest algorithm identifier includes parameters to align with signtool

netmackan:
Solves the difference in signature output as discussed in #139 by making sure that the digest algorithm identifier contains the parameters and if not includes a DER NULL.
With this change the ASN.1 of a signature created with signtool is identical to what was produced by Jsign.

Note that this implementation still uses the DefaultDigestAlgorithmIdentifierFinder in order to support existing and future algorithms with other parameters (today I believe only Ed448 would have a digest algorithm with parameters but may be others). The method is marked protected so that it can be overriden in case it turns out some special case would need some other logic.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/AuthenticodeSigner.java
+++ b/jsign-core/src/main/java/net/jsign/AuthenticodeSigner.java
@@ -54,6 +54,7 @@ import org.bouncycastle.cms.SignerInformationStore;
 import org.bouncycastle.cms.SignerInformationVerifier;
 import org.bouncycastle.cms.jcajce.JcaSignerInfoVerifierBuilder;
 import org.bouncycastle.operator.ContentSigner;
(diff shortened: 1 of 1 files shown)


ebourg:
Thank you, but it looks like it breaks one of the unit tests:

    PESignerTest.testWithSignatureAlgorithmSHA256withRSAandMGF1:542 Digest algorithm expected:<2.16.840.1.101.3.4.2.1> but was:<1.3.14.3.2.26>

netmackan:
I have pushed an update to the PR. I made a mix-up with which digest algorithm to look at. Before it used the configured digestAlgorithm while it should actually use the digest algorithm from the ContentSigner (i.e. being the digest algorithm from the signature algorithm).

ebourg:
Merged, thank you!

PR #142, opened 2023-03-22
Enable SHA-1 signing with Azure KeyVault

drolevar:
Despite the fact that SHA-1 signatures are being deprecated, they are still necessary in some legacy cases.
This PR enables to perform SHA-1 signing with Azure KeyVault.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/jca/AzureKeyVaultSigningService.java
+++ b/jsign-core/src/main/java/net/jsign/jca/AzureKeyVaultSigningService.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.KeyStoreException;
-import java.security.MessageDigest;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
@@ -35,6 +34,9 @@ import java.util.Map;
 import com.cedarsoftware.util.io.JsonWriter;
 
 import net.jsign.DigestAlgorithm;
+import org.bouncycastle.asn1.DERNull;
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
+import org.bouncycastle.asn1.x509.DigestInfo;
 
 /**
  * Signing service using the Azure KeyVault API.
@@ -55,6 +57,7 @@ public class AzureKeyVaultSigningService implements SigningService {
      */
     private final Map<String, String> algorithmMapping = new HashMap<>();
     {
+        algorithmMapping.put("SHA1withRSA", "RSNULL");
         algorithmMapping.put("SHA256withRSA", "RS256");
         algorithmMapping.put("SHA384withRSA", "RS384");
         algorithmMapping.put("SHA512withRSA", "RS512");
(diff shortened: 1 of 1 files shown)


ebourg:
Thank you for the PR. Out of curiosity, what legacy cases do you have in mind? Signing applications for old Windows versions?

Azure requires SHA-1 hashes to be wrapped in a DigestInfo, but not the other digest algorithms? That's odd, there is no mention of that in the [API documentation](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/sign/sign), how did you find out?

ebourg:
I've found this info:

https://learn.microsoft.com/en-us/azure/key-vault/keys/about-keys-details#rsa-algorithms

> The DigestInfo is constructed on the server side for Sign operations that algorithms RS256, RS384 and RS512 generate.

So indeed, for RSNULL, the DigestInfo is constructed on the client side. And probably for the other algorithms too.

ebourg:
Looking at the Azure SDK, the hash is never wrapped in a DigestInfo, even for EC and RSA-PSS signatures:

https://github.com/Azure/azure-sdk-for-java/blob/1228af1b/sdk/keyvault/azure-security-keyvault-jca/src/main/java/com/azure/security/keyvault/jca/implementation/KeyVaultClient.java#L389

So this looks very specific to RSNULL. I wonder if RSNULL could work with other algorithms and not just SHA-1.

jsign-core/src/main/java/net/jsign/jca/AzureKeyVaultSigningService.java
```diff
@@ -152,9 +174,14 @@ public byte[] sign(SigningServicePrivateKey privateKey, String algorithm, byte[]
             throw new InvalidAlgorithmParameterException("Unsupported signing algorithm: " + algorithm);
         }
 
-        MessageDigest digest = DigestAlgorithm.of(algorithm.substring(0, algorithm.toLowerCase().indexOf("with"))).getMessageDigest();
+        DigestAlgorithm digestAlgorithm = DigestAlgorithm.of(algorithm.substring(0, algorithm.toLowerCase().indexOf("with")));
+        MessageDigest digest = digestAlgorithm.getMessageDigest();
         data = digest.digest(data);
 
+        if (digestAlgorithm == DigestAlgorithm.SHA1) {
```
ebourg, line 181: Better check that `"RSNULL".equals(alg)`, since this probably works with any algorithm and not just SHA-1. Also `createSha1RawDigest` could be inlined in the `sign` method.

drolevar:
> This is also supported by AzureSignTool: [vcsjones/AzureSignTool#28](https://github.com/vcsjones/AzureSignTool/issues/28)

It is, but AzureSignTool relies on mssign32 which is a pain in the ...

jsign-core/src/main/java/net/jsign/jca/AzureKeyVaultSigningService.java
drolevar: > Better check that `"RSNULL".equals(alg)`, since this probably works with any algorithm and not just SHA-1. Also `createSha1RawDigest` could be inlined in the `sign` method.

Will do.

drolevar:
> Azure requires SHA-1 hashes to be wrapped in a DigestInfo, but not the other digest algorithms? That's odd, there is no mention of that in the [API documentation](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/sign/sign), how did you find out?

I found out by lots of searching on Github and stackoverflow :)

ebourg:
Merged, thank you!

Issue #150, opened 2023-05-31
ECC broken with AWS KMS


vmal-altium:
Trying to sign with an AWS KMS key with KeySpec `ECC_NIST_P256` results in the error message:
```
jsign: Unknown signature type requested: SHA256WITHECC
```

ebourg:
Correction intégrée, merci pour le coup de main !

PR #151, opened 2023-05-31
ECC broken with AWS KMS

vmal-altium:
Simple fix to solve issue #150.
Tested successfully with a self-signed certificate and a `ECC_NIST_P256` key in AWS KMS.


ebourg:
Squashed and merged in https://github.com/ebourg/jsign/commit/5631c83aaddb8127438254feee83628383218aad

Issue #147, opened 2023-05-18
AWS KMS with service-linked role


vmal-altium:
My requirement is to use the KMS option of jsign from an EC2 instance with an appropriate role, without having to manually specify credentials in the command line.

As a result, I was wondering whether there are specific reasons not to use the AWS SDK rather than API (aside from the extra dependency).  The SDK covers more options in terms of credentials (see [DefaultCredentialsProvider](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/credentials/DefaultCredentialsProvider.html)), and hides the request-signing business.

I have done a prototype with the SDK, which seems to work, and I wanted to get your view on how to take this forward.

ebourg:
I try to avoid the AWS SDK due to its size, the REST API is simple enough to be called directly.

How do you call the API with a service linked role? Can you get an access key and a secret key with the AWS CLI?

ebourg:
The input variables are the endpoint (169.254.169.254 in this example) and the name of the role?

vmal-altium:
169.254.169.254 is a fixed IP address that AWS uses to offer the [metadata service](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html) to all EC2 instances, so that is not going to change, and can be essentially hardcoded.

The role is assigned externally to the EC2 instance (AWS console or so), and defines the permissions it has to interact with AWS services (KMS here). The metadata service (IMDS) can be queried from within the instance to know the currently assigned role name, and from that, credentials to make requests to the other AWS services.

So there is no input variable in this case; credentials are available from the fact that the code is run from within an authorised EC2 instance. I would still offer the possibility to pass credentials through the command line (as now), but if not provided, then it would try to use other means to get credentials.

ebourg:
The implementation has been merged with a few changes. The new `AmazonCredentials` class centralizes the access to the credentials and hides the fetching logic, It supports the EC2 instance credentials and credentials set in the environment variables. It could be extended to also load the credentials from a local `~/.aws/credentials` file as used by the AWS CLI.

Could you test and confirm I haven't broken anything?

vmal-altium:
Sorry for the delay.
I have just tested the 5.0 release, and it's all fine for the AWS credentials from IMDS, as well as detecting mismatches between private/public keys.
Thanks a lot for the mention in the release notes.

ebourg:
Thanks again for the help!

ebourg:
[...]

PR #152, opened 2023-05-31
Query IMDSv2 on AWS EC2 to get credentials for KMS if no storepass parameter

vmal-altium:
This is a solution for #147.
- It does not use AWS Java SDK (too heavy bundle).
- If the `storepass` parameter is not provided, it queries the Instance MetaData Service (IMDS) to get credentials to access AWS KMS.
- I did not reuse the `RESTClient` because it assumes JSON response body, and IMDS explicitly returns text, though it may also sometimes be JSON in that text response...

### Note

If running from within a container hosted on that EC2, the IMDSv2 may not be reachable by default (hop limit set to 1 for the PUT request to obtain the API token). Updating the [hop limit](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/configuring-IMDS-existing-instances.html) to 2 solves the issue, but this is not managed here. Just a clue in case of breaking tests.

### DIFF
--- a/jsign-core/src/main/java/net/jsign/jca/AmazonSigningService.java
+++ b/jsign-core/src/main/java/net/jsign/jca/AmazonSigningService.java
@@ -17,9 +17,7 @@
 package net.jsign.jca;
 
 import java.io.IOException;
-import java.io.InterruptedIOException;
 import java.net.HttpURLConnection;
-import java.net.SocketException;
 import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
@@ -49,6 +47,7 @@ import com.cedarsoftware.util.io.JsonWriter;
 import org.apache.commons.codec.binary.Hex;
 
 import net.jsign.DigestAlgorithm;
+import net.jsign.jca.IMDS2Client.UnreachableServiceException;
 
 import static java.nio.charset.StandardCharsets.*;
 
@@ -112,10 +111,10 @@ public class AmazonSigningService implements SigningService {
         } else {
             try {
                 elements = IMDS2Client.create().getCredentials();
-            } catch (SocketException | InterruptedIOException e) {
+            } catch (UnreachableServiceException e) {
                 throw new IllegalArgumentException("storepass " + parameterName
                         + " must specify the AWS credentials: <accessKey>|<secretKey>[|<sessionToken>]"
-                        + ", when not running from an EC2 instance (IMDSv2 service was unreachable; check the hop limit if containerized)", e);
+                        + ", when not running from an EC2 instance (" + e.getMessage() + ")", e);
             } catch (IOException e) {
                 throw new RuntimeException("an error occurred while fetching temporary credentials from IMDSv2 service", e);
             }
--- a/jsign-core/src/main/java/net/jsign/jca/AmazonSigningService.java
+++ b/jsign-core/src/main/java/net/jsign/jca/AmazonSigningService.java
@@ -17,7 +17,9 @@
 package net.jsign.jca;
 
 import java.io.IOException;
+import java.io.InterruptedIOException;
 import java.net.HttpURLConnection;
+import java.net.SocketException;
 import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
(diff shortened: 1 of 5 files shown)


jsign-core/src/main/java/net/jsign/jca/AmazonSigningService.java
```diff
@@ -91,8 +91,8 @@ public class AmazonSigningService implements SigningService {
     public AmazonSigningService(String region, String credentials, Function<String, Certificate[]> certificateStore) {
         this.certificateStore = certificateStore;
 
-        // parse the credentials
-        String[] elements = credentials.split("\\|", 3);
+        // Obtain the credentials
+        String[] elements = credentials != null ? credentials.split("\\|", 3) : IMDS2Client.create().getCredentials();
         if (elements.length < 2) {
             throw new IllegalArgumentException("Invalid AWS credentials: " + credentials);
```
ebourg, line 97: Here if `credentials` was null and IMDS2Client didn't return the credentials, I think the IllegalArgumentException should reuse the error message removed from `KeyStoreType`

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
```diff
@@ -0,0 +1,146 @@
+package net.jsign.jca;
```
ebourg, line 1: The license header is missing

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Obtain a token to authorize queries to IMDSv2.
+     */
+    private String getApiToken() {
+        if (apiToken != null) // TODO: check token TTL if long-lived client
+            return apiToken;
+        try {
+            URL url = new URL(IMDS_ENDPOINT + "/latest/api/token");
+            HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+            conn.setRequestMethod("PUT");
```
ebourg, line 91: A connect timeout is necessary here, something like 2 or 3 seconds should be enough

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
```diff
@@ -0,0 +1,146 @@
...
+
+import com.cedarsoftware.util.io.JsonReader;
+
+/**
+ * Client to query the Instance MetaData Service (IMDS) v2 from AWS EC2 instances.
+ * 
+ * @since 5.0
+ * @see <a href="https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html">Retrieve instance metadata</a>
+ */
+public class IMDS2Client {
```
ebourg, line 19: The class could be made package private

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private static final String ROLE_PATTERN = "[-\\w+=,.@]{1,64}";
+    private static final String IMDS_ENDPOINT = "http://169.254.169.254";
+    private static final int TOKEN_TTL_SECONDS = 21600; // 6h (default & max value)
+
+    String apiToken = null;
```
ebourg, line 25: `apiToken` could be private

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
vmal-altium: Header added

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
vmal-altium: I've added 3s timeout for the API token and 10s for the other calls.

jsign-core/src/main/java/net/jsign/jca/IMDS2Client.java
vmal-altium: updated to private

jsign-core/src/main/java/net/jsign/jca/AmazonSigningService.java
ebourg: Thank you for the changes. I'll move the credentials into a dedicated class, that will solve the propagation of `parameterName` to `AmazonSigningService`.

vmal-altium: So I went with a custom exception to flag that **AWS** metadata service is not reachable (which I should have done from the start...). Hopefully the error message provides some hint of what goes wrong, for a user to troubleshoot, but it means the message is not identical when running on GitHub Actions (Azure) or locally (I replaced the string comparison in the test with pattern matching).

Now, on Azure, we get:
```
jsign: storepass option must specify the AWS credentials: <accessKey>|<secretKey>[|<sessionToken>], when not running from an EC2 instance (IMDSv2 host did not respond as expected; are you in AWS cloud?)
```
and locally:
```
jsign: storepass option must specify the AWS credentials: <accessKey>|<secretKey>[|<sessionToken>], when not running from an EC2 instance (IMDSv2 host was unreachable; check the hop limit if containerized)
```

ebourg:
PR merged, thank you!

PR #153, opened 2023-06-02
Support passing of passwords via file

cpfeiffer:
This PR adds support for supplying the password for keystore and private key through a file instead of passing them directly as a parameter to the CLI.

Passing them as a parameter makes them visible to everyone who can run `ps -au`, so pretty much to every user on the system. Using files with tight permissions OTOH can make this a bit more secure.

We're having this change for years already and I'm wondering if this would be applicable for your version as well.


ebourg:
Thank you for the PR, but this is already implemented in `KeyStoreBuilder:readPassword‎()` when the password is prefixed with `file:`. It also supports loading from an environment variable with the `env:` prefix.

cpfeiffer:
Ah, thank you, I will try using that instead.

PR #160, opened 2023-06-21
fix for crash with small files

paolodenti:
When reading small files, less than 8 bytes, the size of a long, the `DataInputStream.readLong` fails throwing `EOFException`.


ebourg:
Thanks! Do you think you could add a unit test too?

paolodenti:
Sure, I will

ebourg:
It looks good thank you

Issue #81, opened 2020-07-08
APPX/MSIX file support


g88704:
Sorry beginner question...

I want to sign Microsoft MSIX-packages under Linux. I tried your "jsign" tool but with MSIX-Files I get a "jsign: Unsupported file" error.

- Did I do something wrong or is signing "MSIX"-files not supported/possible?

Thanks in advance and Kind Regards

Michael

ebourg:
Hi Michael, thank you for bringing this to my attention. Jsign doesn't support MSIX files, this could be an improvement for a future version.

ebourg:
Some resources about the MSIX format:
* [MSIX documentation](https://docs.microsoft.com/en-us/windows/msix/)
* [MSIX SDK](https://github.com/microsoft/msix-packaging/)

Sample files:
* [Windows Terminal](https://github.com/microsoft/terminal/releases/download/v1.1.1812.0/Microsoft.WindowsTerminalPreview_1.1.1812.0_8wekyb3d8bbwe.msixbundle) (msix bundle)
* [MSIX Hero](https://msixhero.net/get/) (msix)

Quick overview:
* The format is zip based
* The extension is either `.msix` or `.msixbundle`
* The signature is an entry named `AppxSignature.p7x`, it consists in a PKCS#7 signedData ASN.1 structure prefixed by `PKCX`.
* The spcSipInfoObjID version is 16842752 (0x01010000) and the UUID is `4BDFC50A-07CE-E24D-B76E-23C839A09FD1` for a msix file, and `B3585F0F-DEAA-9A4B-A434-95742D92ECEB` for a bundle.
* The digest in the spcIndirectDataContext structure is a concatenation of several digests. It starts with a header `APPX` and is followed by an array of names (4 bytes) + hash (32 bytes) (see the [DigestName](https://github.com/microsoft/msix-packaging/blob/8395c165960e093f7a2d09fe39ee5a4a9e582d10/src/inc/internal/AppxSignature.hpp#L35) enum).
* The hash algorithm used for the signature must match the algorithm defined in the `AppxBlockMap.xml` entry (as an URI in the `HashMethod` attribute of the root `BlockMap` element, see [MSIX package signing](https://docs.microsoft.com/en-us/windows/msix/package/sign-app-package-using-signtool#using-signtool)).

ebourg:
APPX files used for UWP applications use the same format. The spcSipInfoObjID version and UUID are identical.

[...]

PR #164, opened 2023-06-28
Checks of Signable interface case insensitive

mat1e:
As discussed in #81, the checks of supported extensions in Signable interface is case sensitive but it should not be.

This PR simply makes the checks case insensitive by checking extension with lowercases.


ebourg:
Merci EDF :)

PR #166, opened 2023-07-05
Add project map to README.md

alexkarezin:
Adding a link to the high-level diagrams including module, library dependency and others (https://sourcespy.com/github/ebourgjsign/). Built directly from source and updated on schedule. Intended to simplify developer's introduction to the project.

In the spirit of transparency - I am the author of the diagrams. Hope contributors find it useful.


ebourg:
Thank you for the suggestion. I see the value for complex projects, but Jsign has just a handful of modules, that's not that helpful.

The diagrams could be a bit clearer, some arrows are difficult to follow. Using only horizontal and vertical lines like IntelliJ dependency graph may be a good idea. And a different background for test dependencies is also worth considering.

image

alexkarezin:
Emmanuel, thank you. I really appreciate the feedback. Let me take a last look at the whole set and try to address the concerns you raised. If you don't mind I'm going to ask for your opinion again later. Thanks again for a thorough writeup! @ebourg

PR #206, opened 2024-03-01
Fixes #148 -  log4j stderr message about not finding a logging implementation when using jsign CLI

### LINKED ISSUE #148: Error message: Log4j2 could not find a logging implementation
I tried to upgrade from 3.1 to 4.2.  My build script did well to sign several dlls, exes, and a ps1.  When it got to signing an msi, something went wrong.  But there was apparently a problem outputting the error message, as I got:
`ERROR StatusLogger Log4j2 could not find a logging implementation. Please add log4j-core to the classpath. Using SimpleLogger to log to the console...`
There was nothing before or after it.

It was before outputting anything like "Adding authenticode signature to..."

Reverting back to 3.1 made the problem go away.

davoustp:
This fix makes the suggestion from https://github.com/ebourg/jsign/issues/148#issuecomment-1555797023 permanent, by leveraging the log4j2 properties mechanism (instead of relying onto users to do it themselves, struggling with passing Java props to an executable `jar`).

Reference:
* Factory loading shortcut:
https://github.com/apache/logging-log4j2/blob/9c234901e15c9e7f54e527e56d994aa100bc1cbf/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java#L72-L83

* Properties loading:
https://github.com/apache/logging-log4j2/blob/9c234901e15c9e7f54e527e56d994aa100bc1cbf/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java#L131-L133

* Log4J properties filename:
https://github.com/apache/logging-log4j2/blob/9c234901e15c9e7f54e527e56d994aa100bc1cbf/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java#L61

### DIFF
--- a/jsign/src/choco/tools/jsign.cmd
+++ b/jsign/src/choco/tools/jsign.cmd
@@ -3,5 +3,4 @@
 java %JSIGN_OPTS% ^
      -Djava.net.useSystemProxies=true ^
      -Dbasename=jsign ^
-     -Dlog4j2.loggerContextFactory=net.jsign.log4j.simple.SimpleLoggerContextFactory ^
      -jar %~dp0\jsign.jar %*
--- a/jsign/src/deb/data/usr/share/jsign/jsign.sh
+++ b/jsign/src/deb/data/usr/share/jsign/jsign.sh
@@ -3,5 +3,4 @@
 java  $JSIGN_OPTS \
      -Djava.net.useSystemProxies=true \
      -Dbasename=`basename "$0"` \
-     -Dlog4j2.loggerContextFactory=net.jsign.log4j.simple.SimpleLoggerContextFactory \
      -jar /usr/share/jsign/jsign-@VERSION@.jar "$@"
--- a/jsign/src/main/resources/log4j2.component.properties
+++ b/jsign/src/main/resources/log4j2.component.properties
@@ -1,0 +1,1 @@
+log4j2.loggerContextFactory=net.jsign.log4j.simple.SimpleLoggerContextFactory


ebourg:
Merci ! I guess the `log4j2.loggerContextFactory` property can be removed from `jsign.sh` and `jsign.cmd` too.

davoustp:
> Merci ! I guess the `log4j2.loggerContextFactory` property can be removed from `jsign.sh` and `jsign.cmd` too.

De rien! ;-)
Did this small change as well (these files went under the radar), but had the very bad idea to squash commits then force push - so this latest change it does not appear in this PR... Any way to make this happen?

tresf:
Sorry for necromancing here, but this seems to have created a new issue, I believe due to a signature change in newer log4j versions.  As most loggers errors are -- it's benign -- but it's pretty hard to suppress since now the properties file is embedded into the project.  Catch-22? 😅 

Yes, we can fix this with a JVM param, e.g. `-Dlog4j2.loggerContextFactory=net.jsign.log4j.simple.SimpleLoggerContextFactory` but that introduces exactly what the OP wanted to avoid... passing Java props.

```
2026-04-15T02:08:41.264861Z main ERROR Class net.jsign.log4j.simple.SimpleLoggerContextFactory specified in
 the log4j2.loggerContextFactory system property does not extend org.apache.logging.log4j.spi.LoggerContextFactory
```

ebourg:
@tresf I guess the warning is from your application and not from Jsign CLI, right? If you have net.jsign:jsign and log42 on the classpath that's expected, you should depend on net.jsign:jsign-core instead.

tresf:
> @tresf I guess the warning is from your application and not from Jsign CLI, right? If you have net.jsign:jsign and log42 on the classpath that's expected, you should depend on net.jsign:jsign-core instead.

Yeah that was exactly it.  I lazily placed it with my test libs and now I'm dealing with the consequences.  `jsign-core` fixes the warnings but then my ant tasks get much (much) more complicated (I use both CLI and signjar).  Thanks for the quick sanity check.  I'll ignore it in my tests until I can move it to a dedicated location.

Issue #119, opened 2022-04-19
Add ability to sign with additional certificates


MrAlex94:
Various other signing tools offer the ability to sign with additional certificates:

[SignTool](https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool#sign-command-options):

Sign command option | Description
-- | --
/ac  file | Adds an additional certificate from file to the signature block.

[AzureSignTool](https://github.com/vcsjones/AzureSignTool/#parameters):

Sign command option | Description
-- | --
--additional-certificates [short: -ac, required: no] | A list of paths to additional certificates to aide in building a full chain for the signing certificate. Azure SignTool will build a chain, either as deep as it can or to a trusted root. This will also use the Windows certificate store, in addition to any certificates specified with this option. Specifying this option does not guarantee the inclusion of the certificate, only if it is part of the chain. To include multiple certificates, specify this option mulitple times, such as -ac file1.cer -ac file2.cer. The files specified must be public certificates only. They cannot be PFX, PKCS12 or PFX files.

It would be great if `jsign` also had this ability.

ebourg:
Did you try the `certfile` parameter? This allows you to define the certificates added to the signature.

ebourg:
@laurentgo thank you for the PRs, I'll give them a look. Could you elaborate on the credentials issue with Google KMS?

ebourg:
@laurentgo I've checked the documentation about application default credentials, have you tried getting an access token with `gcloud auth application-default print-access-token` and use it for the `storepass` argument?

https://cloud.google.com/sdk/gcloud/reference/auth/application-default/print-access-token

laurentgo:
I guess there's nothing wrong with `gcloud` command, it's just that our security policies recommend using ADC and not exposing password as arguments, so using GoogleKMS type with jsign is not an option for now, while using `libkmsp11` (which supports ADC) as a PKCS#11 provider for jsign doesn't have this authentication issue (but requires an alternative certificate chain).

One nice thing I discovered is that Google has extensive documentation regarding how ADC is supposed to work from an implementation point of view:  https://google.aip.dev/auth/4110
Now, this is probably hard/time consuming to implement vs simply using google libraries (but I can understand this is probably not something jsign want to do).

ebourg:
Technically the access token is not a password, but if your security policies prohibit using it as a command line argument it's possible to read the token from an environment variable, and the actual value never appears as an argument:

[...]

PR #215, opened 2024-04-11
Add the jsign.certfile system property to feed an additional certificate with the JCA provider

umireon:
Maybe related to #119

## Background

We have developed an Android app and Google requires us to sign our build artifact when we upload it to Google Play Console.
For better security, we'd like to sign our build artifact with Google Cloud KMS and we found this excellent product.
We composed the following command to sign our aab with Cloud KMS.

```
jarsigner -J-cp -Jjsign-6.0.jar -J--add-modules -Jjava.sql \
  -providerClass net.jsign.jca.JsignJcaProvider \
  -providerArg projects/PROJECT_ID/locations/global/keyRings/KEYRING_NAME \
  -keystore NONE \
  -storetype GOOGLECLOUD \
  -storepass "$(gcloud auth print-access-token)" \
  -certchain full-chain.pem \
  ./app/build/outputs/bundle/release/app-release.aab \
  KEY_NAME/cryptoKeyVersions/VERSION
```

However, it does not work with the following error.

```
jarsigner error: java.lang.RuntimeException: Failed to load the certificate from 
```

I read the code of jsign and found that the GOOGLECLOUD provider needs the public key to sign but Cloud KMS does not store certificates.
In this PR, we'd like to propose to add a way to feed an additional certificate to Jsign when it is used as a JCA provider.

## Overview

When Jsign is used as a JCA provider, namely with jarsigner, we can provide additional options via the system property of Java.
Adding this PR enable users to sign JAR artifacts with jarsigner and Cloud KMS by the following command.

```
jarsigner -J-cp -Jjsign-6.1-SNAPSHOT.jar -J--add-modules -Jjava.sql \
  -J-Djsign.certfile=full-chain.pem \
  -providerClass net.jsign.jca.JsignJcaProvider \
  -providerArg projects/PROJECT_ID/locations/global/keyRings/KEYRING_NAME \
  -keystore NONE \
  -storetype GOOGLECLOUD \
  -storepass "$(gcloud auth print-access-token)" \
  -certchain full-chain.pem \
  ./app/build/outputs/bundle/release/app-release.aab \
  KEY_NAME/cryptoKeyVersions/VERSION
```

Thank you in advance!

Best,

### DIFF
--- a/jsign-core/src/main/java/net/jsign/jca/JsignJcaProvider.java
+++ b/jsign-core/src/main/java/net/jsign/jca/JsignJcaProvider.java
@@ -102,7 +102,12 @@ public class JsignJcaProvider extends Provider {
         public JsignJcaKeyStore(KeyStoreType type, String keystore) {
             builder.storetype(type);
             builder.keystore(keystore);
-            builder.certfile("");
+            String certfile = System.getProperty("jsign.certfile");
+            if (certfile == null) {
+                builder.certfile("");
+            } else {
+                builder.certfile(certfile);
+            }
         }
 
         private KeyStore getKeyStore() throws KeyStoreException {


ebourg:
Thank you for the PR, however this is maybe not necessary.

Jsign needs to know the algorithm of the key for signing (RSA of ECDSA), when using Google Cloud there are 3 ways to get this information:
* from the certificate chain provided, however the jarsigner certchain parameter isn't propagated to Jsign
* from the key description, but this requires an extra GCP API call and the `cloudkms.cryptoKeyVersions.get` permission or the _Cloud KMS Viewer_ role
* from the alias if the algorithm is appended, like this: `mykey/cryptoKeyVersions/2:RSA`

Assuming you are using the latest release of Jsign, I think you can fix this issue either by adjusting the permissions or by appending `:RSA` to the key alias. Let me know if this works for you.

umireon:
Thank you for suggestion, I will try later!

umireon:
I got it worked with the RSA suffix!

PR #229, opened 2024-06-06
Improve HashiCorpVaultSigningService to support multiple secret engines and enhance error handling

eatasaur:
- Added support for HashiCorp Vault Transit secrets engine along with the existing Google Cloud KMS.
- Integrate logging using Java's `Logger` to capture warnings and errors.
- Refactor the class to better handle different key types ("algorithm" for GCP KMS and "type" for Transit).
- Introduce boolean flag `googleCloudKMS` to determine the secrets engine type.
- Enhance exception handling to cover various error scenarios including null values and unexpected responses.
- Update unit tests to cover the new logic and ensure correctness in handling both secret engines.
- Suppress unchecked warnings appropriately.


ebourg:
Thank you for the PR. Does the Transit engine store the key in an HSM?

ebourg:
I agree the boolean isn't ideal. I see two solutions:

1. Try to guess the type of the endpoint, either GCP or Transit, by looking for `gcp` or `transit` in the endpoint URI, and if it's inconclusive, call an API method that only works for one of the two engines. For example [`/config`](https://developer.hashicorp.com/vault/api-docs/secret/gcpkms#read-configuration) for GCPKMS or [`/config/keys`](https://developer.hashicorp.com/vault/api-docs/secret/transit#read-keys-configuration) for Transit.
2. Use a different SigningService for each engine. For example `HASHICORPVAULT/TRANSIT` and keep `HASHICORPVAULT` for GCPKMS or rename it to `HASHICORPVAULT/GCPKMS`. Or maybe just `TRANSIT` since Vault has been forked as [OpenBao](https://openbao.org) recently, so me may as well drop the HashiCorp name.

eatasaur:
Initially, I was wanting to add a new SigningService for Vault Transit, but ended up not to, as that was a bit more work, and I'm not fully aware of the overhead associated with allowing `jsign` to support a new SigningService. That would, of course, fulfil the requirement.

That being said, both solutions would be perfectly valid in my opinion, so please let me know how you'd prefer to proceed from here. If there are changes that can be implemented by me, I would try to put in the effort to do so.

Thanks!

ebourg:
Different services for each engine is probably cleaner and easier to implement.

ebourg:
@eatasaur Do you think you could squash your commits and rebase on top of the current master branch please?

jsign-crypto/src/main/java/net/jsign/jca/HashiCorpVaultSigningService.java
```diff
@@ -131,15 +198,81 @@ public byte[] sign(SigningServicePrivateKey privateKey, String algorithm, byte[]
...
+            if (response == null || !response.containsKey("data")) {
+                throw new IllegalArgumentException("Response or 'data' key is null.");
+            }
+
+            if (response.containsKey("warnings") && response.get("warnings") != null) {
+                logger.log(Level.WARNING, String.format("Vault responded with warnings: %s", response.get("warnings")));
+            }
+
+            if (response.containsKey("error")) {
+                throw new IllegalArgumentException(String.format("Vault responded with error: %s", response.get("error")));
```
ebourg, line 236: Are you sure this statement is reached? If the server returns an HTTP error 4xx or 5xx it will be caught by the error handling code in RESTClient.

jsign-crypto/src/main/java/net/jsign/jca/HashiCorpVaultSigningService.java
```diff
@@ -131,15 +198,81 @@ public byte[] sign(SigningServicePrivateKey privateKey, String algorithm, byte[]
...
+            }
 
             return Base64.getDecoder().decode(signature);
+
+        } catch (IllegalArgumentException | ClassCastException e) {
+            logger.log(Level.SEVERE, "Invalid response format: ", e.getMessage());
+            throw new GeneralSecurityException("Invalid response format.", e);
         } catch (IOException e) {
-            throw new GeneralSecurityException(e);
+            logger.log(Level.SEVERE, "IO Exception during signing: ", e.getMessage());
```
ebourg, line 271: No need to log the exception, if it's thrown it will be logged or displayed elsewhere

jsign-core/src/main/java/net/jsign/jca/HashiCorpVaultSigningService.java
```diff
@@ -0,0 +1,283 @@
...
+            try {
+                shaType = digestAlgorithm.toString().split("SHA")[1];
+            } catch (RuntimeException e) {
+                logger.log(Level.WARNING, "Could not determine SHA type - defaulting to SHA256");
+                shaType = "256";
+            }
+
+            request.put("input", Base64.getEncoder().encodeToString(data));
+            request.put("prehashed", true);
+            request.put("hash_algorithm", String.format("sha2-%s", shaType));
```
ebourg, line 218: `String.format("sha2-%s", shaType)` is a bit overkill for a simple concatenation, `"sha2-" + shaType` is ok. Transit doesn't support SHA1?

jsign-crypto/src/main/java/net/jsign/jca/HashiCorpVaultSigningService.java
eatasaur: You are correct. This code would otherwise not be reachable. 
Removed in next commit.

jsign-crypto/src/main/java/net/jsign/jca/HashiCorpVaultSigningService.java
eatasaur: Removed the logging on this scenario in the next commit.

jsign-core/src/main/java/net/jsign/jca/HashiCorpVaultSigningService.java
eatasaur: Makes sense, I've changed it accordingly in the next commit.
The formatting for SHA1 is a bit different from any SHA2 algorithm, as you can see in [HashiCorp Vault Official Documentation - Transit API - Sign Data](https://developer.hashicorp.com/vault/api-docs/secret/transit#sign-data).

ebourg:
I've merged your changes with some simplifications (less logging and error handling to keep the code short, it's still possible to analyze potential errors using the new debug mode). I've noticed that the Transit support could be further improved by loading the certificate associated to the key (using the [export key](https://developer.hashicorp.com/vault/api-docs/secret/transit#export-key) endpoint).

Thank you for your contribution!

PR #245, opened 2024-09-22
feat: support FIPS endpoints for AWS KMS

sbstjn:
AWS services offer FIPS endpoints for some services and regions. For [AWS KMS, all regions support FIPS endpoints](https://docs.aws.amazon.com/general/latest/gr/kms.html#kms_region). When using the CLI, you can normally use `--endpoint-url https://kms-fips.us-east-1.amazonaws.com` and when using any AWS SDK, you can configure the endpoint "mode" by setting `AWS_USE_FIPS_ENDPOINT` in the environment to `true`. 

This change tries to add comperable support for this.

default / current : `kms.us-east-1.amazonaws.com`
fips endpoint: `kms-fips.us-east-1.amazonaws.com`

Details:
https://docs.aws.amazon.com/sdkref/latest/guide/feature-endpoints.html

KMS Service Endpoints:
https://docs.aws.amazon.com/general/latest/gr/kms.html#kms_region

### DIFF
--- a/jsign-crypto/src/main/java/net/jsign/jca/AmazonSigningService.java
+++ b/jsign-crypto/src/main/java/net/jsign/jca/AmazonSigningService.java
@@ -81,6 +81,21 @@ public class AmazonSigningService implements SigningService {
         algorithmMapping.put("SHA512withRSA/PSS", "RSASSA_PSS_SHA_512");
     }
 
+     /**
+     * Generates the endpoint URL for the given AWS region.
+     *
+     * @param region the AWS region
+     * @return the endpoint URL
+     */
+    public static String getEndpointUrl(String region) {
+        String useFipsEndpoint = System.getenv("AWS_USE_FIPS_ENDPOINT");
+        if (useFipsEndpoint != null && useFipsEndpoint.equalsIgnoreCase("true")) {
+            return "https://kms-fips." + region + ".amazonaws.com";
+        }
+        
+        return "https://kms." + region + ".amazonaws.com";
+    }
+
     /**
      * Creates a new AWS signing service.
      *
@@ -90,7 +105,7 @@ public class AmazonSigningService implements SigningService {
      * @since 6.0
      */
     public AmazonSigningService(String region, Supplier<AmazonCredentials> credentials, Function<String, Certificate[]> certificateStore) {
-        this(credentials, certificateStore, "https://kms." + region + ".amazonaws.com");
+        this(credentials, certificateStore, getEndpointUrl(region));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ebourg:
Thank you for the PR, I wasn't aware of the AWS FIPS mode. It would be nice to make the FIPS mode usable even without setting an environment variable. The AmazonSigningService constructor could be modified to accept either a region or a full endpoint URL, or maybe a boolean to enable the FIPS mode.

sbstjn:
I wanted to simulate the default AWS behaviour, as I don't want to have an additional configuration for `jsign` about FIPS and for AWS CLI etc.

jsign-crypto/src/main/java/net/jsign/jca/AmazonSigningService.java
```diff
@@ -259,7 +274,7 @@ void sign(HttpURLConnection conn, AmazonCredentials credentials, byte[] content,
         String host = endpoint.getHost();
         Matcher matcher = hostnamePattern.matcher(host);
         String regionName = matcher.matches() ? matcher.group(2) : "us-east-1";
-        String serviceName = matcher.matches() ? matcher.group(1) : "kms";
+        String serviceName = "kms";
```
ebourg, line 277: I'd rather strip the `-fips` suffix here, so this method remains generic for any AWS service

jsign-crypto/src/main/java/net/jsign/jca/AmazonSigningService.java
```diff
@@ -81,6 +81,21 @@ public class AmazonSigningService implements SigningService {
         algorithmMapping.put("SHA512withRSA/PSS", "RSASSA_PSS_SHA_512");
     }
 
+     /**
+     * Generates the endpoint URL for the given AWS region.
+     *
+     * @param region the AWS region
+     * @return the endpoint URL
+     */
+    public static String getEndpointUrl(String region) {
```
ebourg, line 90: Is it possible to make this method non static and package privarte?

sbstjn: I assumed, if I want to test this individually, it needs to be public. Don't know that much about Java ;) Can give it a try later …

jsign-crypto/src/main/java/net/jsign/jca/AmazonSigningService.java
sbstjn: But that class is only intended to access KMS or are there other services involved that I miss?

jsign-crypto/src/main/java/net/jsign/jca/AmazonSigningService.java
ebourg: It remains testable from a test class in the same package if the public keyword is removed, that's the "package private" visibility.

ebourg:
The PR has been merged with a few minor changes, please let me know if I broke anything. Thank you for the help!

PR #267, opened 2024-12-07
Make `PrivateKeyUtils#load` method file extension agnostic

### LINKED ISSUE #264: Consider implementing magic-based private key file type detection, or a CLI parameter to specify the type
Currently, the `jsign` CLI is hardcoded to infer the format of a private key file from its extension:

https://github.com/ebourg/jsign/blob/6acd186a6a6bac4c917ee7e84c130cabe86e5fe4/jsign-crypto/src/main/java/net/jsign/PrivateKeyUtils.java#L66-L74

This logic is fine in the vast majority of cases, but breaks down when private key files do not have a `.pem` or `.pvk` extension, and there is no switch to manually specify the private key file type in the CLI. PEM files in particular need not have a file extension altogether; this is common practice with SSH PEM keys, for example. And, for security reasons, strict server-side repository push hooks may deny pushes for files with such extensions in order to prevent leaking secrets by accident, forcing e.g. test keys to be marked as such by having different extensions.

It'd be great for my use cases if `jsign` either tried parsing the key with the supported formats in sequence until one succeeds, some magic-based format detection was made, or a CLI switch like `--keytype` was added to specify this.

AlexTMjugador:
As outlined in #264, this is achieved by trying to parse the key file according to one of the supported formats in sequence until one works. Given that there are only two supported formats at the moment, and that PEM files are attempted first and more common in the wild than PVK, this approach should have good enough performance. Because a Java exception can only have a single cause, I've attached the underlying parse exceptions to the higher-level `KeyException` as supressed exceptions. These get displayed to consumers on e.g. stacktraces, allowing them to know what exactly went wrong when parsing either format.

While at it, I've added a test to ensure that this extension-agnostic behavior is maintained over time. Also, the `PrivateKeyUtils.java` file had inconsistent line endings with respect to the rest of the repository, so I fixed that, causing Git to report a big diff there.

Resolves #264.

## Private key load errors display preview

Note how the underlying exceptions are marked as "suppressed", but still retain their message, cause and other diagnostic data.

```
$ java -jar jsign/target/jsign-7.0-SNAPSHOT.jar sign \
  --certfile jsign-core/src/test/resources/keystores/jsign-test-certificate.pem \
  --keyfile jsign-core/src/test/resources/keystores/privatekey.asc \
  jsign-core/src/test/resources/wineyes.exe
jsign: Failed to load the keystore 
java.security.KeyStoreException: Failed to load the private key from jsign-core/src/test/resources/keystores/privatekey.asc
        at net.jsign.KeyStoreType$1.getKeystore(KeyStoreType.java:95)
        at net.jsign.KeyStoreBuilder.build(KeyStoreBuilder.java:285)
        at net.jsign.SignerHelper.build(SignerHelper.java:327)
        at net.jsign.SignerHelper.sign(SignerHelper.java:450)
        at net.jsign.SignerHelper.execute(SignerHelper.java:305)
        at net.jsign.JsignCLI.execute(JsignCLI.java:213)
        at net.jsign.JsignCLI.main(JsignCLI.java:57)
Caused by: java.security.KeyException: Failed to load the private key from jsign-core/src/test/resources/keystores/privatekey.asc (valid PEM or PVK file expected)
        at net.jsign.PrivateKeyUtils.load(PrivateKeyUtils.java:80)
        at net.jsign.KeyStoreType$1.getKeystore(KeyStoreType.java:93)
        ... 6 more
        Suppressed: net.jsign.bouncycastle.util.encoders.DecoderException: unable to decode base64 string: invalid characters encountered in base64 data
                at net.jsign.bouncycastle.util.encoders.Base64.decode(Unknown Source)
                at net.jsign.bouncycastle.util.io.pem.PemReader.loadObject(Unknown Source)
                at net.jsign.bouncycastle.util.io.pem.PemReader.readPemObject(Unknown Source)
                at net.jsign.bouncycastle.openssl.PEMParser.readObject(Unknown Source)
                at net.jsign.PrivateKeyUtils.readPrivateKeyPEM(PrivateKeyUtils.java:119)
                at net.jsign.PrivateKeyUtils.load(PrivateKeyUtils.java:68)
                ... 7 more
        Caused by: java.io.IOException: invalid characters encountered in base64 data
                at net.jsign.bouncycastle.util.encoders.Base64Encoder.decode(Unknown Source)
                ... 13 more
        Suppressed: java.lang.IllegalArgumentException: PVK header signature not found
                at net.jsign.PVK.parse(PVK.java:71)
                at net.jsign.PVK.parse(PVK.java:61)
                at net.jsign.PrivateKeyUtils.load(PrivateKeyUtils.java:75)
                ... 7 more
Try `java -jar jsign.jar --help' for more information.
```

### DIFF
--- a/jsign-crypto/src/main/java/net/jsign/PrivateKeyUtils.java
+++ b/jsign-crypto/src/main/java/net/jsign/PrivateKeyUtils.java
@@ -54,26 +54,33 @@ public class PrivateKeyUtils {
 
     /**
      * Load the private key from the specified file. Supported formats are PVK and PEM,
-     * encrypted or not. The type of the file is inferred from its extension (<code>.pvk</code>
-     * for PVK files, <code>.pem</code> for PEM files).
-     * 
+     * encrypted or not. The type of the file is inferred by trying the supported formats
+     * in sequence until one parses successfully.
+     *
      * @param file     the file to load the key from
      * @param password the password protecting the key
      * @return the private key loaded
      * @throws KeyException if the key cannot be loaded
      */
     public static PrivateKey load(File file, String password) throws KeyException {
+        Exception pemParseException;
         try {
-            if (file.getName().endsWith(".pvk")) {
-                return PVK.parse(file, password);
-            } else if (file.getName().endsWith(".pem")) {
-                return readPrivateKeyPEM(file, password != null ? password.toCharArray() : null);
-            }
+            return readPrivateKeyPEM(file, password != null ? password.toCharArray() : null);
         } catch (Exception e) {
-            throw new KeyException("Failed to load the private key from " + file, e);
+            pemParseException = e;
         }
-        
-        throw new IllegalArgumentException("Unsupported private key format (PEM or PVK file expected");
+
+        Exception pvkParseException;
+        try {
+            return PVK.parse(file, password);
+        } catch (Exception e) {
+            pvkParseException = e;
+        }
+
+        KeyException keyException = new KeyException("Failed to load the private key from " + file + " (valid PEM or PVK file expected)");
+        keyException.addSuppressed(pemParseException);
+        keyException.addSuppressed(pvkParseException);
+        throw keyException;
     }
 
     /**
--- a/jsign-crypto/src/test/java/net/jsign/PrivateKeyUtilsTest.java
+++ b/jsign-crypto/src/test/java/net/jsign/PrivateKeyUtilsTest.java
@@ -19,6 +19,8 @@ package net.jsign;
 import java.io.File;
 import java.io.FileWriter;
 import java.math.BigInteger;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.KeyException;
 import java.security.PrivateKey;
 import java.security.interfaces.ECPrivateKey;
@@ -55,6 +57,14 @@ public class PrivateKeyUtilsTest {
     }
 
     @Test
+    public void testLoadPKCS1PEMNonPEMExtension() throws Exception {
+        File targetFile = new File("target/test-classes/keystores/privatekey.pkcs1.pem.key");
+        Files.copy(new File("target/test-classes/keystores/privatekey.pkcs1.pem").toPath(), targetFile.toPath());
+
+        testLoadPEM(targetFile, null);
+    }
+
+    @Test
     public void testLoadEncryptedPKCS1PEM() throws Exception {
         testLoadPEM(new File("target/test-classes/keystores/privatekey-encrypted.pkcs1.pem"), "password");
     }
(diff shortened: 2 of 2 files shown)


ebourg:
It looks good, thanks a lot. Using suppressed exceptions is as good idea.

I know the line endings are inconsistent, but I'd prefer to keep them as is to keep the diff readable by default. I'll probably normalize the whole project at some point in the future anyway.

jsign-core/src/test/resources/keystores/generate-private-key.sh
```diff
@@ -8,6 +8,7 @@ openssl genrsa -aes128 -traditional -passout pass:password -out privatekey-encry
 openssl pkcs8 -topk8       -in privatekey-encrypted.pkcs1.pem -out privatekey-encrypted.pkcs8.pem -passin pass:password -passout pass:password -v2 aes128
 openssl pkcs8 -topk8       -in privatekey-encrypted.pkcs1.pem -out privatekey.pkcs8.pem -passin pass:password -nocrypt
 openssl pkcs8 -traditional -in privatekey.pkcs8.pem           -out privatekey.pkcs1.pem -nocrypt
+cp privatekey.pkcs1.pem privatekey.pkcs1.pem.key
```
ebourg, line 11: No need to copy the file here, this can be done in the test method

jsign-crypto/src/main/java/net/jsign/PrivateKeyUtils.java
```diff
@@ -1,154 +1,161 @@
...
-     * for PVK files, <code>.pem</code> for PEM files).
-     * 
-     * @param file     the file to load the key from
-     * @param password the password protecting the key
-     * @return the private key loaded
-     * @throws KeyException if the key cannot be loaded
-     */
-    public static PrivateKey load(File file, String password) throws KeyException {
-        try {
-            if (file.getName().endsWith(".pvk")) {
```
ebourg, line 67: What about keeping the extension based detection and then falling back to the other method?

jsign-core/src/test/resources/keystores/generate-private-key.sh
AlexTMjugador: Good point, I think that's indeed cleaner. I'm pushing a commit to do that. Thanks!

jsign-crypto/src/main/java/net/jsign/PrivateKeyUtils.java
AlexTMjugador: I considered that approach for a while, and realized it wouldn't handle cases where key files have misleading extensions (e.g., a PVK key with a `.pem` extension, or vice versa). Ultimately, I believe that allowing the key to be parsed regardless of the file extension, even if it's misleading, is more user-friendly. That said, I don’t feel strongly about this and am open to changing it if requested :smile:

ebourg:
That's fine, let's keep it this way. Could you just squash your commits into one before I merge the PR please?

AlexTMjugador:
Okay! I've just squashed the commits into one :+1:

.gitignore
```diff
@@ -2,6 +2,7 @@ target/
 /jsign-gradle-plugin/.gradle
 /jsign-gradle-plugin/build
 dependency-reduced-pom.xml
+/jsign-core/src/test/resources/keystores/privatekey.pkcs1.pem.key
```
ebourg, line 5: One last thing, the file should be copied in the `target/test-classes/keystores` directory, it will be removed automatically by `mvn clean` and there is no need to declare it in `.gitignore`

AlexTMjugador: Right, this is a leftover from when the shell script copied the file. I've amended my squashed commit to remove this :+1:

ebourg:
Perfect, thank you!

PR #274, opened 2025-01-29
doc: pkcs11 info

zosocanuck:
Add Venafi CodeSign Protect as a PKCS#11 provider and provide some basic documentation on how to leverage PKCS#11 for signing.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -703,7 +703,7 @@ the name of the one to use.</p>
 The <code>alias</code> represents the name of the certificate used for signing in the <code>keystore</code>.
 
 <pre>
- jsign --storetype PKCS11 --storepass NONE --keystore pkcs11.conf --alias SIGNATURE application.exe
+ jsign --storetype PKCS11 --storepass NONE --keystore pkcs11.conf --alias My-Signing-Cert application.exe
 </pre>
 
 <h4 id="example-awskms">Signing with AWS Key Management Service</h4>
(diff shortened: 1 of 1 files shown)


ebourg:
Thank you for the PR, I agree the PKCS#11 documentation could be improved.

Regarding Venafi I think there are too many PKCS#11 implementations to list them all, I prefer mentioning a few of the most popular hardware based ones.

However it looks like Venafi CodeSign Protect is a cloud signing service, it could be integrated directly into Jsign without using the PKCS#11 module. Do you know if its API is documented?

zosocanuck:
Venafi CodeSign Protect is currently only a self-hosted solution with a well-documented API.  That said it may be much easier to use the PKCS#11 integration approach with a roadmap item to integrate natively via API.  Thoughts?

ebourg:
I think I prefer the API integration, PKCS#11 is a pain to use.

I got a look at the documentation, the REST API seems pretty straightforward:

https://docs.venafi.com/Docs/24.1API/#?route=post-/vedhsm/api/sign

zosocanuck:
Sounds good. Let me know how I can provide help with the API integration given that I work at Venafi (A CyberArk Company).

ebourg:
If you want to implement it I can guide you through the process. There are several examples in the jsign-crypto module. Otherwise I'd need a temporary access to a CodeSign Protect instance.

zosocanuck:
I can work on getting this implemented and will open a separate PR.  It would still be good to include an example of how to use the PKCS#11 keystore.

ebourg:
I've implemented a new signing service for SignPath (5a4418562414dd71080d0ee74c59c95e135dcd4c), you can use this commit as a template for CodeSign Protect.

PR #285, opened 2025-02-14
Add Scoop package to docs

sitiom:
I have just added [`jsign`](https://scoop.sh/#/apps?q=jsign&id=48a014807579e3b45a673676f26a17fff1f8e961) to Scoop: https://github.com/ScoopInstaller/Main/pull/6137. It would be nice to have a reference to it on the docs.


ebourg:
Thank you very much for the package. The link added to the documentation contains a session id, it can probably be removed.

sitiom:
It's not a session ID; it seems to be a unique identifier that the website generates that is linked to a specific package.

ebourg:
The same URL without the id seems to work though.

PR #286, opened 2025-02-15
Add support for Venafi CodeSign Protect

zosocanuck:
This PR adds support for Venafi CodeSign Protect.


ebourg:
Thank you very much for the PR, I'll review it

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,269 @@
...
+                }
+            } catch (IOException e) {
+                throw new KeyStoreException("Unable to retrieve the Venafi keystore with alias: " + alias, e);
+            }
+        }
+    }
+
+    @Override
+    public List<String> aliases() throws KeyStoreException {
+        return new ArrayList<>(certificates.keySet());
```
ebourg, line 142: Is there an API call to initialize that could be used to initialize the certificates available? Otherwise this method will return an empty list if `getCertificateChain()` hasn't been called yet.

jsign-crypto/src/main/java/net/jsign/jca/VenafiCredentials.java
```diff
@@ -0,0 +1,72 @@
+/**
```
ebourg, line 1: I changed the license headers two days ago, they now start with `/*` instead of `/**`

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,269 @@
...
+           
+            data = digestAlgorithm.getMessageDigest().digest(data);
+            out.write(data);
+            byte[] arr_combined = out.toByteArray();
+
+            Map<String, Object> request = new HashMap<>();
+
+            Map<String, String> clientInfo = new HashMap<>();
+            clientInfo.put("ClientLibraryName", "jsign");
+            clientInfo.put("ClientLibraryVersion", "7.0.0");
```
ebourg, line 223: Is the version absolutely required? I'll certainly forget to update it.

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,269 @@
...
+        algorithmMapping.put("SHA384withECDSA", 4165);
+        algorithmMapping.put("SHA512withECDSA", 4166);
+    }
+
+    /* Map ASN.1 DER prefix structures to MessageDigest Algorithm */
+
+    private final byte[] getHashPrefix(Integer mechanism) {
+        switch (mechanism) {
+            case 64: case 4164:
+                return new byte[]{0x30, 0x31, 0x30, 0x0d, 0x06, 0x09, 0x60, (byte)0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x01, 0x05, 0x00, 0x04, 0x20}; // SHA256
```
ebourg, line 82: Couldn't we use BouncyCastle to generate the ASN.1 structure instead of assembling the bytes manually?

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,269 @@
...
+        } catch (KeyStoreException e) {
+            throw (UnrecoverableKeyException) new UnrecoverableKeyException().initCause(e);
+        }
+
+    }
+
+    public byte[] encodeASN1(byte[] sigBytes) throws IOException {
+
+        // Split the sigbytes into r and s components
+        BigInteger r = new BigInteger(1, java.util.Arrays.copyOfRange(sigBytes, 0, sigBytes.length / 2));
```
ebourg, line 182: `import java.util.Arrays` is preferred over a fully qualified class name

jsign-crypto/src/test/java/net/jsign/KeyStoreBuilderTest.java
```diff
@@ -305,6 +305,38 @@ public void testBuildGaraSign() throws Exception {
         assertNotNull("keystore", keystore);
     }
 
+    @Test
+    public void testBuildVenafi() throws Exception {
+        KeyStoreBuilder builder = new KeyStoreBuilder().storetype(VENAFI);
+
+        Exception e = assertThrows(IllegalArgumentException.class, builder::build);
+        assertEquals("message", "storepass parameter must specify the Venafi username/password and/or the path to the keystore containing the TLS client certificate: <username>|<password>, <certificate>, or <username>|<password>|<certificate>", e.getMessage());
```
ebourg, line 313: Does this test really pass? The `validate()` method doesn't handle the third parameter with the client certificate.

jsign-crypto/src/test/java/net/jsign/jca/VenafiSigningServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+    public void tearDown() {
+        closeJadler();
+    }
+
+    @Test
+    public void testGetAliases() throws Exception {
+        SigningService service = new VenafiSigningService("http://localhost:" + port(), new VenafiCredentials("username", "password", null));
+        List<String> aliases = service.aliases();
+
+        assertEquals("aliases", Collections.emptyList(), aliases);
```
ebourg, line 53: This doesn't compile, `java.util.Collections` wasn't imported.

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,284 @@
...
+        algorithmMapping.put("SHA256withECDSA", 4164);
+        algorithmMapping.put("SHA384withECDSA", 4165);
+        algorithmMapping.put("SHA512withECDSA", 4166);
+    }
+
+    /* Add ASN.1 DER prefix to MessageDigest */
+
+    public static byte[] addASN1Prefix(DigestAlgorithm digestAlgorithm, byte[] hash) throws IOException {
+        ASN1EncodableVector v = new ASN1EncodableVector();
+        System.out.println(digestAlgorithm.name());
```
ebourg, line 89: `System.out.println()` forgotten here

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,284 @@
...
+        switch(digestAlgorithm.name()) {
+            case "SHA256":
+                v.add(new ASN1ObjectIdentifier("2.16.840.1.101.3.4.2.1"));
```
ebourg, line 92: The OIDs are already in the `DigestAlgorithm` class, you can simply use `digestAlgorithm.oid` and remove the switch.

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+    }
+
+    @Override
+    public SigningServicePrivateKey getPrivateKey(String alias, char[] password) throws UnrecoverableKeyException {
+        try {
+            Certificate[] chain = getCertificateChain(alias);
+            String algorithm = chain[0].getPublicKey().getAlgorithm();
+            System.out.println("algorithm: " + algorithm);
```
ebourg, line 192: `System.out.println()` left here.

jsign-crypto/src/test/java/net/jsign/jca/VenafiSigningServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void testSignWithFailure() throws Exception {
+        onRequest()
+                .havingMethodEqualTo("POST")
+                .havingPathEqualTo("/vedhsm/api/sign")
+                .havingHeaderEqualTo("Authorization", "Bearer da49LbfHokLO+fKwFIneJg==")
+                .respond()
+                .withStatus(200)
+                .withBody(new FileReader("target/test-classes/services/venafi-sign-error.json"));
```
ebourg, line 117: `venafi-sign-error.json` is missing from the PR

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,284 @@
...
+        Integer clientMechanism = algorithmMapping.get(algorithm);
+        if (clientMechanism == null) {
+            throw new InvalidAlgorithmParameterException("Unsupported signing algorithm: " + clientMechanism);
+        }
+
+        try {
+            DigestAlgorithm digestAlgorithm = DigestAlgorithm.of(algorithm.substring(0, algorithm.toLowerCase().indexOf("with")));
+            data = digestAlgorithm.getMessageDigest().digest(data);
+            
+            byte[] arr_combined = getASN1Prefix(digestAlgorithm, data);
```
ebourg, line 233: `getASN1Prefix(DigestAlgorithm, byte[])` doesn't exist.

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,284 @@
...
+        ASN1OutputStream asn1OutputStream = ASN1OutputStream.create(byteArrayOutputStream);
+        asn1OutputStream.writeObject(components);
+        asn1OutputStream.close();
+
+        return byteArrayOutputStream.toByteArray();
+    }
+
+    @Override
+    public byte[] sign(SigningServicePrivateKey privateKey, String algorithm, byte[] data) throws GeneralSecurityException {
+        //System.out.println(algorithm);
```
ebourg, line 223: cleanup

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,284 @@
...
+            String algorithm = chain[0].getPublicKey().getAlgorithm();
+            System.out.println("algorithm: " + algorithm);
+            return new SigningServicePrivateKey(alias, algorithm, this);
+        } catch (KeyStoreException e) {
+            throw (UnrecoverableKeyException) new UnrecoverableKeyException().initCause(e);
+        }
+
+    }
+
+    public byte[] encodeASN1(byte[] sigBytes) throws IOException {
```
ebourg, line 200: What is this ASN.1 structure? Does it have a name?

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,272 @@
...
+           
+            Map<String, ?> response = client.post("/vedhsm/api/sign", JsonWriter.format(request));
+            String status = (String) response.get("Error");
+            if (status != null) {
+                throw new IOException("Signing operation failed: " + response.get("Error"));
+            }    
+            String signature = (String) response.get("ResultData");
+
+            if (algorithm.equals("SHA256withECDSA") || algorithm.equals("SHA384withECDSA") || algorithm.equals("SHA512withECDSA")) {
+                return encodeASN1(Base64.getDecoder().decode(signature));
```
ebourg, line 255: `encoreASN1(byte[])` doesn't exist

ebourg:
Thank you for the quick fixes, but I'm a bit concerned that it doesn't compile. Did you really test it against a real instance of Venafi CodeSign Protect?

zosocanuck:
I just compiled and packaged and didn't run into any issues, and yes I'm testing in a local development environment against a test instance of Venafi.

jsign-crypto/src/test/resources/services/venafi-sign.json
```diff
@@ -0,0 +1,13 @@
+{
```
ebourg, line 1: The json files should contains the API responses, if I'm not mistaken this looks like a request.

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,272 @@
...
+        algorithmMapping.put("SHA384withRSA", 65);
+        algorithmMapping.put("SHA512withRSA", 66);
+        algorithmMapping.put("SHA256withECDSA", 4164);
+        algorithmMapping.put("SHA384withECDSA", 4165);
+        algorithmMapping.put("SHA512withECDSA", 4166);
+    }
+
+    /* Add ASN.1 DER prefix to MessageDigest */
+
+    public static byte[] addASN1Prefix(DigestAlgorithm digestAlgorithm, byte[] hash) throws IOException {
```
ebourg, line 87: This looks like a standard `DigestInfo` structure, it's easier to use the class provided by Bouncy Castle:

    return new DigestInfo(new AlgorithmIdentifier(digestAlgorithm.oid, DERNull.INSTANCE), hash);

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,272 @@
...
+    /** The credentials to authenticate with the service */
+    private final VenafiCredentials credentials;
+
+     /** The name of the Venafi CodeSign Protect certificate KeyId */
+     private String KeyId;
+
+    /** Mapping between Java and Venafi CodeSign Protect signing algorithms */
+    private final Map<String, Integer> algorithmMapping = new HashMap<>();
+    {
+        algorithmMapping.put("SHA256withRSA", 64);
```
ebourg, line 77: For completeness we could add SHA1withRSA (6) and SHA1withECDSA (4162)

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,272 @@
...
+            clientInfo.put("ClientLibraryName", "jsign");
+
+            Map<String, String> processInfo = new HashMap<>();
+            clientInfo.put("Executable", "jsign");
+
+            request.put("ClientInfo", clientInfo);
+            request.put("ProcessInfo", processInfo); 
+            request.put("KeyId", KeyId);
+            request.put("Data", Base64.getEncoder().encodeToString(arr_combined));
+            request.put("ClientMechanism", clientMechanism);
```
ebourg, line 235: The `ClientMechanism` field seems to be optional, does it still work if removed?

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,272 @@
...
+            switch (algorithm) {
```
ebourg, line 236: It might be easier to test the type of key (RSA or ECDSA) instead of enumerating the signature algorithms

jsign-crypto/src/test/java/net/jsign/jca/VenafiSigningServiceTest.java
```diff
@@ -0,0 +1,125 @@
...
+                .withBody(new FileReader("target/test-classes/services/venafi-keystore.json"));
+
+        SigningService service = new VenafiSigningService("http://localhost:" + port(), new VenafiCredentials("username", "password", null));
+
+        Exception e = assertThrows(KeyStoreException.class, () -> service.getCertificateChain("jsign"));
+        assertEquals("message", "Unable to retrieve Venafi certificate 'jsign'", e.getMessage());
+    }
+
+    @Test
+    public void testSign() throws Exception {
```
ebourg, line 91: Could you also add a test covering signing with ECDSA please?

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
zosocanuck: Ok, I added this but at this point the mapping would be no longer used.

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,264 @@
...
+
+            Map<String, String> processInfo = new HashMap<>();
+            clientInfo.put("Executable", "jsign");
+
+            request.put("ClientInfo", clientInfo);
+            request.put("ProcessInfo", processInfo); 
+            request.put("KeyId", KeyId);
+            request.put("Data", Base64.getEncoder().encodeToString(arr_combined));
+            //switch (algorithm) {
+            switch (algorithm.substring(algorithm.toLowerCase().indexOf("with") + 4)) {
```
ebourg, line 228: Or simply `privateKey.getAlgorithm()`

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
```diff
@@ -0,0 +1,264 @@
...
+            }
+           
+            Map<String, ?> response = client.post("/vedhsm/api/sign", JsonWriter.format(request));
+            String status = (String) response.get("Error");
+            if (status != null) {
+                throw new IOException("Signing operation failed: " + response.get("Error"));
+            }    
+            String signature = (String) response.get("ResultData");
+
+            if (algorithm.equals("SHA256withECDSA") || algorithm.equals("SHA384withECDSA") || algorithm.equals("SHA512withECDSA")) {
```
ebourg, line 246: Here too:

```
if ("ECDSA".equals(privateKey.getAlgorithm()) {
    return encodeASN1Signature(Base64.getDecoder().decode(signature));
} else {
    return Base64.getDecoder().decode(signature);
}
```

jsign-crypto/src/main/java/net/jsign/jca/VenafiSigningService.java
ebourg: If the mapping is no longer used it can be removed

jsign-crypto/src/test/java/net/jsign/jca/VenafiSigningServiceTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public void testGetAliases() throws Exception {
+        SigningService service = new VenafiSigningService("http://localhost:" + port(), new VenafiCredentials("username", "password", null));
+        List<String> aliases = service.aliases();
+
+        assertEquals("aliases", Collections.emptyList(), aliases);
+    }
+
+    @Test
+    public void testGetCertificateChain() throws Exception {
+        onRequest()
```
ebourg, line 59: The test server should also mock the requests to /vedauth/authorize/oauth, otherwise the authentication fails due to a 404 error.

zosocanuck: Can you provide more guidance?

ebourg: For each test it's necessary to mock the authentication request by adding something like this:

```
onRequest("/vedauth/authorize/oauth")
    .havingMethodEqualTo("POST")
    .respond()
    .withStatus(200)
    .withBody(new FileReader("target/test-classes/services/venafi-authorize.json"));
```

This will also require a new JSON file in `jsign-crypto/src/test/resources/services/` containing the server response to this API call.

ebourg:
Thank you, there are still some changes necessary but I'll take care of them.

Could you please provide the JSON response to the `/vedhsm/api/getobjects` request for your RSA  test key?

Ideally, if you could import the Jsign test keys and certificates in Venafi and use them to capture the API responses that would be awesome. The keys are in the [test resources directory](https://github.com/ebourg/jsign/tree/master/jsign-core/src/test/resources/keystores):

RSA: https://github.com/ebourg/jsign/blob/master/jsign-core/src/test/resources/keystores/keystore.p12
EC: https://github.com/ebourg/jsign/blob/master/jsign-core/src/test/resources/keystores/keystore-ec.p12

The password is "password".

ebourg:
Alternatively, do you think I could access a Venafi test environment?

zosocanuck:
Here is the `vedhsm/api/getobjects` response for the test RSA key.
[venafi-rsa-getobjects.json](https://github.com/user-attachments/files/18875057/venafi-rsa-getobjects.json)

zosocanuck:
Let me know what else is missing or any other tests that need to be added and I'll get it done 😃

zosocanuck:
@ebourg just curious if there was anything missing or requires an update.

ebourg:
@zosocanuck The application id "VenafiCodeSignClient" is hardcoded in the VenafiCredentials class. Shouldn't this be user configurable? For example with an extra parameter to the storepass parameter, something like this:

```
 jsign --storetype VENAFI \
       --keystore https://example.tpp.local \
       --storepass "<application>|<username>|<password>" \
       --alias my-certificate-label \
       application.exe
```

zosocanuck:
@ebourg I was trying to balance the amount of configuration with out-of-the-box defaults in the Venafi platform so that is why this is hardcoded.  This is a standard client ID that ships with the product and there is a preference to avoid the customization of API integration client IDs for code signing. My preference is to keep as is.

ebourg:
It looks like _Venafi CodeSign Protect_ is now _CyberArk Code Sign Manager_.

https://www.cyberark.com/resources/product-datasheets/cyberark-code-sign-manager

I've contacted the support to request a test access, I'd like to check the implementation.

PR #306, opened 2025-07-12
Find libraries on linux via LD_LIBRARY_PATH

JohnAZoidberg:
It's the common way to find libraries if they're not in the standard /usr/lib, /usr/local/lib, ... directories.

I've tested with yubikey only.

> LD_LIBRARY_PATH=foo/ jsign --storetype YUBIKEY --storepass 123456 --tsaurl http://ts.ssl.com --tsmode RFC3161 foo.exe


JohnAZoidberg:
I'm packaging jsign for NixOS, which does not put libraries in the standard locations.

JohnAZoidberg:
Awesome, thanks for the quick feedback and merging!

PR #308, opened 2025-07-14
Add documentation about how to run from source

JohnAZoidberg:


docs/index.html
```diff
@@ -1117,6 +1117,16 @@ <h4>APK signing</h4>
       --in application.apk
 </pre>
 
+<h3 id="credits">Development</h3>
```
ebourg, line 1120: The id attribute has to be changed

PR #315, opened 2025-08-06
Keep single self-signed certificate in chain

crnzgm:
In version 6.0 I could sign a package with a self signed certificate, whose chain consists only of itself. Since [3bc43b3](https://github.com/ebourg/jsign/commit/3bc43b3a522b1a8f5fdb242ba739b87560d70406) self signed certificates get removed from the chain which causes the signing to fail. To mitigate that, this PR removes self-signed certificates from the chain only if the chain consists of more than 1 certificate.

### DIFF
--- a/jsign-core/src/test/java/net/jsign/PESignerTest.java
+++ b/jsign-core/src/test/java/net/jsign/PESignerTest.java
@@ -627,4 +627,25 @@ public class PESignerTest {
             }
         }
     }
+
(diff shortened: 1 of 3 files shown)


ebourg:
Thank you for the PR. Do you think you could add a unit test too?

crnzgm:
Nevermind, I think I figured it out.

PR #333, opened 2025-11-24
Functional Implementation of the 2023-06-15-preview API (+ Adaptive API changes via ENV variable override)

michael-sabrnak-swi:
This PR was raised in response to https://github.com/ebourg/jsign/issues/332.

Here’s a quick before/after rundown between Azure Trusted Signing’s 2022-style flow and the `2023-06-15-preview` version:

**Endpoint shape**
- _2022_: Traditional REST resource paths like `.../certificateprofiles/{profile}/sign` (no colon).
- _2023_: RPC-style `.../certificateprofiles/{profile}:sign` for the initial POST, while the poll endpoint moves to `/sign/{operationId}` and is advertised via headers. More details [here](https://github.com/Azure/azure-rest-api-specs/blob/a569cbee809be8da35805444742938d7c9ab97b6/specification/trustedsigning/data-plane/TrustedSigning/preview/2023-06-15-preview/azure.developer.trustedsigning.json#L54).

**Operation metadata**
- _2022_: Operation IDs typically lived in the JSON body; headers weren’t essential.
- _2023_: Azure expects clients to read the operation-location header (or the embedded _result.operationId_) and use it to poll. Upgraded _RESTClient_ to expose headers so the service can reliably extract that metadata.

**Response contract**
- _2022_: Many implementations assumed the first response already contained signature data or a simple success/failure.
- _2023_: The first call almost always returns `InProgress`, with the real signature only available after polling the operation URL until `status == "Succeeded"`; errors now come in the `errorDetail`/`errors` envelope.

**Payload subtleties**
- _2022_: Mixed hashing fields (digest + optional hash lists) and some legacy algorithms.
- _2023_: Focuses on base64 digests (`digest`, optional additional hash lists) plus stricter algorithm names (RS256, ES256, etc.) - the refactor clamps inputs to what the preview API accepts.

**Client behavior**

- _2022_: Jsign just needed basic POST/GET calls with JSON parsing.
- _2023_: Requires richer HTTP handling (headers, raw response body, poll timeouts), which is why _RESTClient_ adjusted, updated `AzureTrustedSigningService`, and reworked tests to mimic that asynchronous lifecycle.


michael-sabrnak-swi:
After further investigation, I'm closing this PR. I verified that Microsoft's own latest tooling still uses api-version=`2022-06-15-preview` across the board:                         
                                                                                                                                                                                       
  - **Azure.CodeSigning.Sdk 0.1.164** (Nov 2025) - sole ServiceVersion enum value: V2022_06_15_preview                                                                                     
  - **Microsoft.Trusted.Signing.Client 1.0.95** (bundled in signtool dlib) - same                                                                                                          
  - **Az.CodeSigning 0.2.2 PowerShell module** (Mar 2025) - same                                                                                                                           
  - **Azure.Developer.ArtifactSigning.CryptoProvider 0.1.72** (Mar 2026) - depends on the same SDK                                                                                         

Every downstream tool - the GitHub Action (azure/artifact-signing-action), Azure DevOps tasks (ArtifactSigning 1.0.0, TrustedSigning 0.5.13), the TrustedSigning PowerShell module 0.5.8, and the Notation plugin - all transitively use this same chain. The DLL string tables show REST-style /sign paths (not :sign RPC), and the only API version literal embedded is 2022-06-15-preview.

No newer data plane API version has been published in any Microsoft SDK, and the data plane OpenAPI spec is still not in azure-rest-api-specs (the link referenced in this PR returns 404).

The original issue (#332) was actually caused by a trailing slash in the endpoint URL producing `//codesigningaccounts/`... double-slashes, which started returning 404 after Microsoft tightened URL normalization - not by an API version problem.

ebourg:
Thank you for the info, I'll probably apply the PR when the new API is more widely used.

ebourg:
Reopening since it seems the `2022-06-15-preview` version is now rejected (#350)

michael-sabrnak-swi:
Hi @ebourg, I've made a few tweaks that hopefully satisfy the needs raised in both surendratammineni/jsign#1 and [ebourg/jsign#350](https://github.com/ebourg/jsign/issues/350#issuecomment-4866992451) (comment). The rationale is that we want to be prepared for the worst-case scenario and have some "adaptive" behavior: switching from the default (the `2022-06-15-preview` version Microsoft is still using) to the [stable](https://github.com/Azure/azure-rest-api-specs/tree/3afe65f7239e0e1f26202c75d960266adddb1eeb/specification/artifactsigning/data-plane/ArtifactSigning/stable/2024-06-15) `2024-06-15` version (RPC-style) whenever needed via simple environment variable.

Formal details:

## Make the Azure Trusted Signing API version configurable

### What this changes
- You can now choose which **Azure Trusted Signing (ATS) API version** jsign uses.
- Two ways to set it:
  - Environment variable: `JSIGN_AZURE_API_VERSION=2024-06-15`
  - Constructor argument: `new AzureTrustedSigningService(endpoint, token, "2024-06-15")`
- If you set nothing, jsign behaves exactly as before (nothing breaks).
### How it picks the version
1. Constructor argument (if given)
2. `JSIGN_AZURE_API_VERSION` environment variable (if set)
3. Default: `2022-06-15-preview` (the old behavior)
### Why we did it
Azure changed the shape of the signing API:
- **Old versions** (`2022-06-15-preview`) use a flat REST call: `POST .../certificateprofiles/{profile}/sign`.
- **Newer versions** (`2023-06-15-preview` and later, including the **stable** `2024-06-15`) use an RPC-style call: `POST .../certificateprofiles/{profile}:sign` and return the result nested under a `result` object.
jsign was hard-wired to the old layout, so it could not talk to the newer/stable API. This change lets jsign use either layout depending on the version you pick.
### How the switch works (simple version)
- jsign compares the version string. Anything **`2023-06-15` or later** uses the new RPC-style `:sign` path and parses the nested response.
- Anything older uses the legacy `/sign` path and the flat response.
- Because Azure version strings are dates (`YYYY-MM-DD`), comparing them as text also compares them in time order.
### Forward-compatible on purpose
When Microsoft ships a future version (e.g. `2026-xx-xx`), you don't need a new jsign build. As long as the new version keeps the RPC-style shape, just set `JSIGN_AZURE_API_VERSION=2026-xx-xx` and jsign will already route it correctly. Not a guarantee it works, but it's ready to test on day one.
### This also solves surendratammineni/jsign#1
That PR only swapped the hard-coded version string and kept the old flat parsing. With the stable API, the response is nested, so that approach would misread the reply. Our change is a superset: it supports **both** layouts and lets the user choose the version, so it fixes the same problem in a cleaner, future-proof way.
### Tested against real Azure Trusted Signing
Signed real `.exe` files end-to-end (Azure sign + RFC 3161 timestamp) with all three versions, all succeeded:
| Version | Path used | Result |
|---|---|---|
| `2022-06-15-preview` (default) | legacy `/sign` | Succeeded |
| `2023-06-15-preview` | RPC `:sign` | Succeeded |
| `2024-06-15` (stable) | RPC `:sign` | Succeeded |
Verified with `osslsigncode verify` (valid signature, correct signer certificate) and with jsign's `--debug` output showing the exact URL/verb per version.
### Notes for reviewers
- Default behavior is unchanged; this is fully backward compatible.
- Unit tests cover both the legacy and RPC-style paths.
- No changes to the build workflow are included in this PR.

michael-sabrnak-swi:
Hi @ebourg, is there a realistic chance [this](https://github.com/ebourg/jsign/pull/333#issuecomment-4981340138) could be merged by the end of H2 2026? Happy to address any review feedback or adjust the approach if needed. Thanks for your time.

PR #339, opened 2026-02-25
Skip LDAP URLs in certificate issuer resolution

caglareker:
Fixes #338

Certificates can include Authority Information Access (AIA) extensions with multiple URLs for fetching issuer certificates, including LDAP URLs. The current implementation was attempting to handle all protocols, resulting in an "Unknown protocol ldap" error.

This change adds a test to verify that `CertificateUtils.getIssuerCertificateURL()` correctly skips LDAP URLs and returns HTTP(S) URLs instead. The fix filters out unsupported protocols and uses the first available HTTP URL for issuer certificate retrieval.

### DIFF
--- a/jsign-crypto/src/test/java/net/jsign/CertificateUtilsTest.java
+++ b/jsign-crypto/src/test/java/net/jsign/CertificateUtilsTest.java
@@ -52,6 +52,16 @@ public class CertificateUtilsTest {
     }
 
     @Test
+    public void testGetIssuerCertificateURLSkipsLDAP() throws Exception {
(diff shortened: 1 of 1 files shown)


ebourg:
Thank you for the PR. At some point I plan to refresh the test certificates and remove `jsign-test-certificate-full-chain-2026.pem`, so `testGetIssuerCertificateURLSkipsLDAP()` will become redundant with `testGetIssuerCertificateURL()`. For now the LDAP case is already covered by `testGetFullCertificateChain()`.

PR #346, opened 2026-05-12
Fix #345 : allows keystore of types ETOKEN|PKCS11 to be reloaded more than once until an alias is successfully fetched.

### LINKED ISSUE #345: Empty keystore with SafeNet eToken 5110+ FIPS accessed through PKCS#11 (eTPKCS11.dll)
Hi,

Consider following configuration:

```xml
<!-- Using net.jsign:jsign-maven-plugin:7.4 -->
<configuration>
    <storetype>PKCS11</storetype>
    <keystore>${project.basedir}/digicert-pkcs11.cfg</keystore>
    <replace>true</replace>
    <storepass/>
    <alias>My Existing Alias</alias>
    <tsaurl>http://timestamp.digicert.com</tsaurl>
    <algorithm>SHA-256</algorithm>
</configuration>
```

with cfg:
```cfg
name = eToken
library = C:/Windows/System32/eTPKCS11.dll
slotListIndex = 1
showInfo = false
```

It seems like 8 out of 10 times in run mode, this fails with `No certificate found in the keystore SunPKCS11-eToken`.

After running with breakpoints, I spotted the issue here :
https://github.com/ebourg/jsign/blob/d18f06d4fe24eee93e36f5be8edafd20a20155d0/jsign-crypto/src/main/java/net/jsign/KeyStoreType.java#L697

In case of loading failure, the only difference is the field `((P11KeyStore)ks.keyStoreSpi).aliasMap` being empty instead of size=1. The token info is always here.

It always succeeds (`[INFO] Adding Authenticode signature to C:\Users\codesigning\IdeaProjects\jsign\javaw.exe`) when I stop at this line, which makes me think it's a race condition, the try-with-resource seems not enough to guarantee the keystore has been properly loaded from the pkcs11 token.

Is there any recommandation or known issue to workaround this limitation ? Is it related to the eTPKCS11.dll dll ? Otherwise I planned to fork the project just to retry the ks.load more than 1 time which I'd hate to...

NB: running the same cfg file over `keytool` though **always** works and gets the alias and its certificate properly 
```shell
keytool -list -keystore NONE -storetype PKCS11 -providerclass sun.security.pkcs11.SunPKCS11 -providerArg .\digicert-pkcs11.cfg -v -alias "My Existing Alias"
```

Thank you.

Macadoshis:
Fix #345

### DIFF
--- a/jsign-crypto/src/main/java/net/jsign/KeyStoreBuilder.java
+++ b/jsign-crypto/src/main/java/net/jsign/KeyStoreBuilder.java
@@ -282,7 +282,18 @@ public class KeyStoreBuilder {
      */
     public KeyStore build() throws KeyStoreException {
         validate();
-        return storetype().getKeystore(this, provider());
+        KeyStore keystore = storetype().getKeystore(this, provider());
+        int maxAttempts = storetype().getAliasReloadMaxAttempts();
+        while (!keystore.aliases().hasMoreElements() && maxAttempts-- > 0) {
+            try {
+                Thread.sleep(500);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new KeyStoreException(e);
+            }
+            keystore = storetype().getKeystore(this, provider());
+        }
+        return keystore;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ebourg:
The patch looks fine, thank you. Could you modify it slightly to:
* Use the same max attempt value for all storetypes, I don't think we'll need to customize it per storetype. The value can be hardcoded in the `build()` method.
* document in the `build()` method why the retry mechanism is necessary, with a reference to the related GitHub Issue.

PR #361, opened 2026-09-03
Retry flaky connections

SimonSchwendele:
When using [Azure Artifact Signing] (azure trusted signing) you're currently ewxperiencing degarded performance.  
The api responds slower with each file until it jsign eventually exits.

Eventually we reach `AuthenticodeSignedDataGenerator.getSignerInfo` where where azts dies with a 500 that stops the remaining signatures to be done.  

Our action crashes like this (sorry for the screenshot, our logs expire pretty fast)
image

This extends the currently implemented retry loop to retry on errors as well not just on timeouts.  
I assume that the remotes would rather send 4xx responses when issues arise with the files or this cli.

[Azure Artifact Signing]: https://learn.microsoft.com/en-us/azure/artifact-signing/

### DIFF
--- a/jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
+++ b/jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
@@ -265,7 +265,6 @@ class RESTClient {
                     wait = getBackoffDelayMs(attempt);
                 }
                 log.fine(String.format("HTTP error %d from %s, retrying in %d ms (attempt %d of %d)", responseCode, url, wait, attempt, retries));
-                conn.disconnect();
                 pause(wait);
             } catch (SocketTimeoutException e) {
                 if (attempt >= retries) {
--- a/jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
+++ b/jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
@@ -51,7 +51,7 @@ class RESTClient {
     /** Number of attempts for a request */
     private int retries = 5;
 
-    /** Initial pause between retries in milliseconds, doubled after each attempt on error (not on timeouts) */
+    /** Initial pause between retries in milliseconds, base of the exponential backoff on error (fixed pause on timeouts) */
     private int retryWait = 1000;
 
     /** Connect timeout in milliseconds */
@@ -238,8 +238,9 @@ class RESTClient {
 
     /**
      * Opens a connection to the specified URL and makes several attempts if a timeout occurs
-     * or if the server returns a transient error (rate-limit, bad gateway, etc.).
-     * The pause between the attempts grows exponentially unless the server specifies a <code>Retry-After</code> header.
+     * or if the server returns a retryable error (rate-limit, bad gateway, etc.).
+     * The pause between the attempts grows exponentially up to the connect timeout,
+     * unless the server specifies a <code>Retry-After</code> header.
      * The provided configurator is used to set up the connection before making the request.
      */
     private HttpURLConnection open(URL url, HttpURLConnectionHandler configurator) throws IOException {
@@ -251,7 +252,7 @@ class RESTClient {
                 configurator.handle(conn);
 
                 int responseCode = conn.getResponseCode();
-                if (!isProbablyFlakinessError(responseCode) || attempt >= retries) {
+                if (!isRetryableError(responseCode) || attempt >= retries) {
                     return conn;
                 }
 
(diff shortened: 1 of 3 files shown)


jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
```diff
@@ -237,37 +237,79 @@ public RESTClient readTimeout(int readTimeout) {
...
+                    return conn;
+                }
+
+                Long serverDelay = getServerRetryDelayMs(conn);
+                long wait;
+                if (serverDelay != null) {
+                    wait = serverDelay;
+                    log.fine("Server returned 'RetryAfter': " + serverDelay);
                 } else {
+                    wait = (long) (retryWait * Math.pow(2, attempt - 1));
```
SimonSchwendele: We should probably put a cap (30s?) in here

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
```diff
@@ -237,37 +237,79 @@ public RESTClient readTimeout(int readTimeout) {
...
+                log.fine(String.format("HTTP error %d from %s, retrying in %d ms (attempt %d of %d)", responseCode, url, wait, attempt, retries));
+                conn.disconnect();
+                pause(wait);
+            } catch (SocketTimeoutException e) {
+                if (attempt >= retries) {
                     throw (IOException) new SocketTimeoutException("Unable to connect to " + url + " after " + retries + " attempts").initCause(e);
                 }
+
+                log.fine(String.format("Connection to %s timed out, retrying in %d ms (attempt %d of %d)", url, retryWait, attempt, retries));
+                pause(retryWait);
```
SimonSchwendele, line 275: I don't use the backoff here as timeouts should probably not navigate us into rate limits

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
```diff
@@ -237,37 +237,79 @@ public RESTClient readTimeout(int readTimeout) {
...
             }
+            attempt++;
         }
+    }
 
-        return conn;
+    private boolean isProbablyFlakinessError(int responseCode) {
```
ebourg: I'd rather name the method `isTransientError()` than `isProbablyFlakinessError()`

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
ebourg: I agree

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
ebourg: Yes, it could be capped to the value of the connect timeout.

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
ebourg: Or isRetryableError()

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
ebourg: The AWS SDK uses a 20 sec limit, and a random delay within an exponentially increasing range:

https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html#backoff-formula

    delay = random(0, 1) × min(20,000 ms, base_delay × 2^retry)

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
```diff
@@ -237,37 +237,87 @@ public RESTClient readTimeout(int readTimeout) {
...
+                Long serverDelay = getServerRetryDelayMs(conn);
+                long wait;
+                if (serverDelay != null) {
+                    wait = serverDelay;
+                    log.fine("Server returned 'RetryAfter': " + serverDelay);
                 } else {
+                    wait = getBackoffDelayMs(attempt);
+                }
+                log.fine(String.format("HTTP error %d from %s, retrying in %d ms (attempt %d of %d)", responseCode, url, wait, attempt, retries));
+                conn.disconnect();
```
ebourg: disconnect() is not necessary here, if there is a keep-alive the connection can be reused for the next attempt. Java will disconnect automatically afterward.

ebourg:
Thank you for the PR. Do you think you could add some tests to RESTClientTest ?

jsign-crypto/src/main/java/net/jsign/jca/RESTClient.java
SimonSchwendele: sorry felt like it'd be more clean that way. 
I'll alter that when I am done with work

SimonSchwendele:
> Thank you for the PR. Do you think you could add some tests to RESTClientTest ?

Sure no problem I'll do that after work.

ebourg:
PR merged with a few cosmetic changes, thank you for the help!

PR #362, opened 2026-09-22
reject extra data after the signature in the PE certificate table

itecz26:
**PE certificate table accepts data smuggled past the signature**
the digest excludes the whole certificate table region declared by the data directory, but nothing verifies that the region only holds the signature, so appending bytes and enlarging the certificate table's directory size leaves the digest unchanged and the signature still verifies (CVE-2013-3900 shape). getCertificateTable now rejects a non-EFI table whose declared size runs past the 8-byte-aligned signature entry.


ebourg:
Thank you for the PR. If the goal is to break on signatures rejected by Windows, then maybe this verification should be implemented as a new rule for the SignatureVerifier class. Otherwise it cripples the API and makes it impossible to read the signature from the file.

itecz26:
makes sense, the throw in getCertificateTable was too blunt. moved it into a CertificateTableRule in SignatureVerifier so the signature is still readable and only the verification fails when the certificate table declares more bytes than the signature entry uses. getCertificateTable stays as it was, and PEFile just exposes getCertificateTableTrailingBytes() for the rule to consult. pushed the change with the tests updated to match.

